PR #56, opened 2026-03-28
feat: add apiKey support for provider models fetching

tuxtof:
This pull request updates the way API keys are handled when fetching provider models, allowing the front-end to pass an API key to the backend, which then forwards it as an Authorization header to the provider. This enables secure, per-request authentication for fetching model lists from providers that require an API key.

**API key forwarding for provider model fetching:**

* The backend (`internal/apiserver/server.go`) now reads an `apiKey` query parameter and, if present, adds it as a Bearer token in the `Authorization` header when proxying model list requests to providers. [[1]](diffhunk://#diff-b42035eced7f44e48ded70c3f3a8783b4ee0966b04560c04879260ef5da47672R2601) [[2]](diffhunk://#diff-b42035eced7f44e48ded70c3f3a8783b4ee0966b04560c04879260ef5da47672L2620-R2629)

**Front-end support for API key parameter:**

* The API utility (`web/src/lib/api.ts`) and the hook for fetching model lists (`web/src/hooks/use-model-list.ts`) are updated so that the API key can be optionally passed from the front-end to the backend, and included in the request if available. [[1]](diffhunk://#diff-d349a3bcde67ad21d3c34ba8d691c300d469d0db8d0bb6257c8e2c23ac71d223L66-R67) [[2]](diffhunk://#diff-af490ef10b2e4da7f8055bd843c240828a9e8bcd6e6244b55888fe13f8ce8f7aL842-R844) [[3]](diffhunk://#diff-d349a3bcde67ad21d3c34ba8d691c300d469d0db8d0bb6257c8e2c23ac71d223L116-R116)

### DIFF
--- a/internal/apiserver/server.go
+++ b/internal/apiserver/server.go
@@ -2598,6 +2598,7 @@ func (s *Server) proxyProviderModels(w http.ResponseWriter, r *http.Request) {
 	}
 
 	provider := r.URL.Query().Get("provider")
+	apiKey := r.URL.Query().Get("apiKey")
 
 	// Determine the models endpoint URL.
 	modelsURL := ""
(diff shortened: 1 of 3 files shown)


tuxtof:
> hey nice addition! one small thing tho - the apiKey in the query string could end up in server/proxy logs, would be better to pass it as a request header (like X-Provider-Api-Key) from the frontend to the backend so we dont accidentally leak secrets.

good remark let me propose another PR

tuxtof:
done with #63

PR #66, opened 2026-03-29
fix: update RBAC rules to include metrics.k8s.io permissions for skill sidecars

tuxtof:
My changes in #55 are blocking skill RBAC creation

This pull request updates RBAC (Role-Based Access Control) configurations to grant permissions for accessing Kubernetes metrics via the `metrics.k8s.io` API. This is important for enabling features like resource usage monitoring (e.g., using `kubectl top`) for nodes and pods, especially for skill sidecars.

RBAC permission updates for metrics:

* Added rules in `charts/sympozium/templates/rbac.yaml` to allow `get` and `list` verbs on `nodes` and `pods` resources in the `metrics.k8s.io` API group.
* Added similar rules in `config/rbac/role.yaml` to grant `get` and `list` permissions on `nodes` and `pods` resources in the `metrics.k8s.io` API group.

### DIFF
--- a/charts/sympozium/templates/rbac.yaml
+++ b/charts/sympozium/templates/rbac.yaml
@@ -114,6 +114,10 @@ rules:
   - apiGroups: ["apiextensions.k8s.io"]
     resources: ["customresourcedefinitions"]
     verbs: ["get", "list", "watch"]
+  # Metrics API — needed to grant metrics.k8s.io permissions to skill sidecars (e.g. kubectl top)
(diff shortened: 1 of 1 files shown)


AlexsJones:
You're right — the design doc was wrong, not the code. The controller uses a scoped ClusterRole, not cluster-admin. I've just fixed the design doc to reflect this accurately (9ad717d).

Since we're keeping the scoped RBAC approach (more controlled than cluster-admin), this PR is valid and needed. Will review and merge.

PR #130, opened 2026-04-29
Enhancement: adding volumes to sensible sympozium resources

SamuZad:
Hi @AlexsJones, I love the project!

I was trying to integrate a sympozium agent Vault Secrets Store CSI for enhanced security and ease of use, but noticed that the CRDs don't support the declaration of "arbitary" volumes, which is required to make CSI drivers work. This PR adds the ability to have `volumes` and `volumeMounts` for the following resources:
- Ensemble
- Agent
- AgentRun
- SkillPack

I wasn't entirely certain that SkillPack should be included, but I thought I'd add it for consistency, and on balance I thought it'd be better to have the additional flexibility

Please let me know if this aligns with your vision, or if you'd like me to change anything (like how and where I declare `volumeMounts`)


AlexsJones (commented):
Thanks for the contribution @SamuZad  great spot on the CSI volume gap! One ask before merging: please add a log warning when a user-supplied volume/mount is silently dropped for colliding with a reserved name (`isReservedVolumeName`), so users aren't left guessing why their mount didn't appear.

three-foxes-in-a-trenchcoat (approved):
LGTM — clean implementation. Volumes/volumeMounts fields land in the right CRD types, propagation chain is consistent across ensemble→agent→agentrun→spawner, and the reserved-name collision protection is sensible. Tests cover the key paths well.

One minor note (same as @AlexsJones): adding a `log.V(2).Info` or similar when a user-supplied volume/mount is silently dropped for a reserved-name collision would help with debugging. Not a blocker though — can follow up.

Ship it when @SamuZad squashes the feedback or merges as-is.

SamuZad:
@AlexsJones thank you for the prompt response!

1. I added warning logging and squashed the commits
2. However, I then decided to add further validation, to reduce the changes of "volume weirdness" or unexpected behaviour

PR #134, opened 2026-04-29
Enhancement: CRDs helm chart

SamuZad:
The built-in crd installation of helm [does not upgrade CRDs](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/)

> There is no support at this time for upgrading or deleting CRDs using Helm. This was an explicit decision after much community discussion due to the danger for unintentional data loss. Furthermore, there is currently no community consensus around how to handle CRDs and their lifecycle. As this evolves, Helm will add support for those use cases.

One of the [recommended](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-2-separate-charts) ways to get around this is to create a 2nd chart, explicitly to install-manage CRDS

A lot of projects manage CRDs this way, for instance all of the [rancher charts](https://github.com/rancher/charts/tree/dev-v2.14/charts). I think it would make sense here as well, especially since the CRDs are likely to evolve, and this will help us avoid either uninstalling or manually upgrading the manifests

Of course, I kept the old way of installing CRDs in, and only pushed this as an advanced installation mode to make it convenient for people to experiment and get started quickly

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -135,11 +135,15 @@ jobs:
           ref: main
           token: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Sync CRDs to Helm chart
+      - name: Sync CRDs to Helm charts
         run: |
-          echo "Syncing CRDs to Helm chart..."
+          echo "Syncing CRDs to charts/sympozium/crds/..."
           mkdir -p charts/sympozium/crds
           cp config/crd/bases/*.yaml charts/sympozium/crds/
+          echo "Syncing CRDs to charts/sympozium-crds/templates/..."
+          mkdir -p charts/sympozium-crds/templates
+          rm -f charts/sympozium-crds/templates/sympozium.ai_*.yaml
+          cp config/crd/bases/*.yaml charts/sympozium-crds/templates/
 
       - name: Commit and push
         run: |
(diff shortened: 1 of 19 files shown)


three-foxes-in-a-trenchcoat (commented):
Needs work: CI failing

SamuZad:
> Needs work: CI failing

The CI is no longer failing

AlexsJones (commented):
nice one — this is the right pattern, matches how rancher/cert-manager do it. keeping the original `crds/` dir so single-chart installs still work is a good call.

couple things:

- `Chart.yaml` has `version: 0.10.8` but main is at `0.10.11` now — rebase + `make helm-sync` should sort it, the CRD templates will be behind HEAD otherwise
- `getting-started.md` now has `--skip-crds` on the main chart install, which makes sense for the two-chart flow, but someone who skips the crds chart and just does the main one will end up with no CRDs. might be worth a note or just drop `--skip-crds` from the quick start and only mention it in the advanced section
- the uninstall ordering thing (crds chart deleted first = cascade-deletes all CRs) is inherent to the pattern but worth calling out more prominently, maybe a ⚠️ in the NOTES.txt

CI is green now so the earlier comment about it failing is resolved. lgtm once the version skew is fixed

SamuZad:
Ok, done

AlexsJones:
The `sympozium-crds` chart version is `0.10.8` but main is at `0.10.12`. This should be synced before merge:

```yaml
# charts/sympozium-crds/Chart.yaml
version: 0.10.12
appVersion: 0.10.12
```

Release-please will bump both going forward, but the initial version needs to match.

Everything else looks good — lint passes, CRD content is in sync, and the PR doesn't impact the CLI installer.

PR #144, opened 2026-05-01
fix: restrict WhatsApp channel to self-chat messages only

### LINKED ISSUE #138: Bug: WhatsApp responds to messages outside of self-chat
Had an agent connected channel via WhatsApp. The issue I encountered is that it would process any message sent by me. Including in conversations with other people. I believe the differentiating factor might be that the chat was 1 to 1 with one other person, ie technically not a group but also not only myself.

Claude thinks something like this should help

```
  --- a/channels/whatsapp/main.go
  +++ b/channels/whatsapp/main.go
  @@ -200,7 +200,10 @@ func (wc *WhatsAppChannel) handleInboundMessage(evt *events.Message) {
  -     // Only process messages from the device owner (self-chat mode).
  -     // Skip status broadcasts, group chats, and messages from other people.
  +     // Only process self-chat messages (user messaging themselves).
  +     // Skip broadcasts, groups, and messages the user sends to other people.
        if evt.Info.Chat.Server == "broadcast" {
                return
        }
        if !evt.Info.IsFromMe {
                return
        }
  +     if evt.Info.Chat.String() != evt.Info.Sender.String() {
  +             return
  +     }
```

AlexsJones:
## Summary
- `IsFromMe` is true for messages the user sends in **any** chat, including 1:1 conversations with other people
- Added a `Chat.User != Sender.User` guard so only messages in the user's self-chat are processed
- Uses the `User` field (phone number) rather than full JID `String()` to avoid `@s.whatsapp.net` vs `@lid` server suffix mismatches

Closes #138

## Test plan
- [ ] Connect WhatsApp channel, send a message in self-chat — should be processed
- [ ] Send a message in a 1:1 chat with another person — should be ignored
- [ ] Send a message in a group chat — should be ignored (existing guard)

### DIFF
--- a/channels/whatsapp/main.go
+++ b/channels/whatsapp/main.go
@@ -197,12 +197,19 @@ func (wc *WhatsAppChannel) handleInboundMessage(evt *events.Message) {
 		evt.Info.Sender.String(), evt.Info.Chat.String(), evt.Info.IsFromMe, evt.Info.IsGroup, evt.Info.Chat.Server,
 		truncateText(extractText(evt.Message), 50))
 
-	// Only process messages from the device owner (self-chat mode).
-	// Skip status broadcasts, group chats, and messages from other people.
+	// Only process self-chat messages (user messaging themselves).
+	// Skip broadcasts, groups, and messages the user sends to other people.
 	if evt.Info.Chat.Server == "broadcast" {
 		return
 	}
 	if !evt.Info.IsFromMe {
+		return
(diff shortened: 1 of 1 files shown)


three-foxes-in-a-trenchcoat:
@AlexsJones Please run `cargo fmt` to format the code.

three-foxes-in-a-trenchcoat (commented):
LGTM

PR #146, opened 2026-05-01
fix(web): prevent gateway toggle from disconnecting UI

Juwebien:
## Summary
- disables the Gateway on/off mutation when the dashboard is reached through the cluster gateway
- keeps copy/read-only config visible so users understand GitOps owns preview Gateway state
- prevents clicking the UI button from setting gateway.enabled=false and deleting the route serving the UI

## Validation
- npm ci
- npm run -s build

## OpsClaw preview context
Clicking the current Gateway toggle in the WebUI can make ui.sympozium.opstech.dev inaccessible because the API writes SympoziumConfig.gateway.enabled=false, the controller deletes Gateway/HTTPRoute resources, and the UI cuts off its own public path. Preview Gateway state is GitOps-owned and should not be mutable from the page.

### DIFF
--- a/web/src/pages/gateway.tsx
+++ b/web/src/pages/gateway.tsx
@@ -76,6 +76,14 @@ export function GatewayPage() {
 
   const handleSave = async () => {
     const isNew = !data?.phase;
+    if (data?.enabled && !form.enabled) {
+      toast.error(
+        "This WebUI is served through the Gateway. Disabling it from here would make the UI unreachable. Use GitOps or kubectl for an intentional edge teardown.",
+      );
+      setForm((prev) => ({ ...prev, enabled: true }));
+      setDirty(false);
+      return;
+    }
     try {
       if (isNew) {
         await createMutation.mutateAsync(form);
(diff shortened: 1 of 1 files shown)


three-foxes-in-a-trenchcoat (commented):
Needs work: CI unknown

AlexsJones (commented):
Nice catch — this is a real foot-gun and I'm glad someone picked it up. Disabling the gateway from the page it serves is exactly the kind of thing that bites you in prod at 2am. The intent here is solid and the UX copy explaining *why* is helpful.

A few suggestions to tighten it up:

**1. The `handleSave` guard (lines 79-86) is dead code now**
The onClick handler already prevents `form.enabled` from ever being set to `false`, so the save-time check can never fire. I'd remove it to avoid confusion — the onClick guard alone is sufficient.

**2. `setDirty(false)` in the handleSave guard could discard unrelated edits**
If a user changed baseDomain or TLS settings and *then* hit save while enabled was still true (the only possible state), this wouldn't fire. But if somehow it did, it would silently reset dirty and swallow their other changes. Another reason to just remove this block.

**3. Button label "On (GitOps-managed)" might confuse users**
The gateway isn't necessarily GitOps-managed — a user could have enabled it through this very page. Something like just `"On"` with the helper text below is enough context. The helper text already explains the teardown story.

**4. Toast severity mismatch**
`toast.warning` on button click vs `toast.error` on save for the same scenario. Since the handleSave guard should go away, this resolves itself, but worth noting.

**5. Minor — helper text could be more concise**
The explanation is good but a bit long for inline form text. Something like:
> *"The WebUI is served through this Gateway. Disable via kubectl or GitOps to avoid disconnecting this page."*

Overall this is a good defensive UX change — just needs a small cleanup pass. 👍

three-foxes-in-a-trenchcoat (commented):
## Review

**Verdict: LGTM**

Good, focused fix. The safety guard for the gateway toggle makes total sense — preventing self-disconnect is a real UX win. The explanatory text in the UI helps users understand why the action is restricted.

One tiny nit: the `setDirty(false)` call in handleSave might prevent the "unsaved changes" indicator from working as expected if there are other form fields modified. Worth a quick check, but not blocking.

three-foxes-in-a-trenchcoat (approved):
## Review

**Verdict: LGTM**

Nice fix — the gateway toggle was indeed a footgun for preview environments. The approach of disabling the mutation when accessed through the cluster gateway is clean, and keeping the read-only config visible so users still understand GitOps ownership is good UX.

All CI checks pass. Ready to merge.

PR #148, opened 2026-05-02
fix(web-proxy): dedupe retried chat requests

Juwebien:
## Summary
- add request hashing for web-proxy chat completions, using Idempotency-Key when present and prompt/model fingerprint otherwise
- reuse a recent matching AgentRun instead of creating duplicate model calls during transport/client retry loops
- add unit coverage for stable fingerprints and recent-run selection

## Validation
- PATH=/home/linuxbrew/.linuxbrew/bin:$PATH GOROOT=$(/home/linuxbrew/.linuxbrew/bin/go env GOROOT) go test ./internal/webproxy

## Ops note
This is the quota guard for Alfy/Sympozium transport failures where the client can retry while the first AgentRun still burns provider tokens.

### DIFF
--- a/internal/webproxy/openai.go
+++ b/internal/webproxy/openai.go
@@ -2,14 +2,18 @@ package webproxy
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"net/http"
 	"os"
+	"sort"
 	"strings"
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	sympoziumv1alpha1 "github.com/sympozium-ai/sympozium/api/v1alpha1"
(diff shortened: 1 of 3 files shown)


three-foxes-in-a-trenchcoat (commented):
Good fix for the retry token-burn issue. A few observations:

-  does a full  on AgentRun objects. This should be fine at current scale, but if you add a label index on  in the controller eventually it'll save you.
- Hash truncation to 16 hex chars (64 bits) is fine for a 15-min TTL window, but worth noting collision is theoretically possible under very high throughput.
- The gateway UI fix is solid — preventing self-disconnection is a real UX win.

LGTM, approved.

three-foxes-in-a-trenchcoat (commented):
Good fix for the retry token-burn issue. A few observations:

- findRecentWebRun does a full List on AgentRun objects. Should be fine at current scale, but consider adding a label index on "sympozium.ai/request-hash" in the controller eventually.
- Hash truncation to 16 hex chars (64 bits) is fine for a 15-min TTL window. Collision is theoretically possible under very high throughput.
- The gateway UI fix is solid - preventing self-disconnection is a real UX win.

LGTM, approved.

three-foxes-in-a-trenchcoat (approved):
## Review

**Verdict: LGTM**

Excellent feature. The request hashing approach is smart — using `Idempotency-Key` when present and falling back to a prompt/model fingerprint otherwise handles both explicit retry-safe headers and transparent client retries.

The TTL-based deduplication (15 min) is a reasonable default. The unit tests are thorough: they cover idempotency-key dominance, stable fingerprinting, newest-match selection, and TTL expiry.

One suggestion: consider making the TTL configurable per-instance rather than hardcoded. Some workloads might benefit from shorter or longer dedup windows. But that's a nice-to-have, not a blocker.

PR #153, opened 2026-05-03
Enhancement: Channel Pod CSI Compatibility

SamuZad:
This PR ensures the compatibility of channel deployments, with the following 2 enhacements:
1. ensures that the channel pods have their own dedicated service account
    - it makes sense not to have the same SA as the agent pods, but I think they shouldn't use `default`, either
 2. in case the pod uses a CSI volume "relax" the requirement of the "channel credential secret" needing to exist prior to creating the deployment

### DIFF
--- a/internal/controller/agent_controller.go
+++ b/internal/controller/agent_controller.go
@@ -38,7 +38,7 @@ type AgentReconciler struct {
 // +kubebuilder:rbac:groups=sympozium.ai,resources=agents/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=sympozium.ai,resources=agents/finalizers,verbs=update
 // +kubebuilder:rbac:groups=apps,resources=deployments,verbs=get;list;watch;create;update;patch;delete
-// +kubebuilder:rbac:groups="",resources=secrets;configmaps;services,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups="",resources=secrets;configmaps;services;serviceaccounts,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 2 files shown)


AlexsJones (commented):
## Review

Clean PR — good use case (Vault CSI integration), solid test coverage, minimal blast radius. Two items to address and a couple of minor suggestions.

### Issues

**1. Orphaned ServiceAccount (low-medium)**

`ensureChannelServiceAccount` creates `sympozium-channel` without an owner reference, so it's never garbage-collected if all Agents in the namespace are deleted. This is probably fine — it's namespace-scoped infrastructure — but worth a comment in the code noting the intentional design choice so future readers don't try to "fix" it.

**2. `channelMountsCSI` is broad (low)**

Any CSI volume on the channel skips the secret existence check, even CSI volumes unrelated to the `configRef` Secret (e.g. a CSI-backed scratch volume). In practice this is pragmatic since users mounting CSI on channel pods are almost certainly doing Vault/secret injection, but a code comment explaining the assumption would help.

### Suggestions

- The hardcoded SA name `"sympozium-channel"` appears in 3 places (`ensureChannelServiceAccount`, `buildChannelDeployment`, and the test). Consider extracting to a `const channelServiceAccountName` to avoid drift.

- Make sure `make manifests` is run before merge — the `+kubebuilder:rbac` marker adding `serviceaccounts` needs to be regenerated into the ClusterRole YAML.

### Tests

Thorough coverage — `channelMountsCSI` helper, secret-gating for both CSI and non-CSI, SA ensure/noop, and SA wiring in `buildChannelDeployment`. Well structured.

PR #158, opened 2026-05-06
Enhancement: add optional target to execute_command IPC requests

SamuZad:
# Problem

There are a number of examples with multiple `skillPacks` defined, for example in a recent [comment](https://github.com/sympozium-ai/sympozium/issues/47#issuecomment-4322505959):
Screenshot 2026-05-06 at 18 11 01

However, when an agent has multiple SkillPacks with sidecars (e.g. `github-gitops` + 
`something-else`), all sidecars share `/ipc/tools/` and race via 
`mkdir .claim-<id>` to handle each `execute_command` request. There's no
mechanism in the protocol to address a specific sidecar. When sidecars have
different capabilities (e.g. one has `gh` + `GH_TOKEN`, and the other does not),
commands can, and will land land in the wrong container and fail with
"command not found"..

## Fix
Additive, backwards-compatible:
- `execRequest` JSON gains an optional `target` field.
- `execute_command` tool advertises an optional `target` parameter.
- Controller injects `SYMPOZIUM_SKILL_PACK=<name>` into each skill sidecar 
  and `SYMPOZIUM_SKILL_TARGETS=<csv>` into the agent container.
- `tool-executor.sh` skips requests whose target doesn't match its 
  `SYMPOZIUM_SKILL_PACK` (case-insensitive, whitespace-trimmed).
- Empty target → today's mkdir-claim race is preserved (legacy behaviour).
- Agent-runner lowercases + trims target before writing the request.

However, it's possible that I misunderstood something/did not read the code correctly

### DIFF
--- a/cmd/agent-runner/tools.go
+++ b/cmd/agent-runner/tools.go
@@ -40,10 +40,13 @@ func defaultTools() []ToolDef {
 	return []ToolDef{
 		{
 			Name: ToolExecuteCommand,
-			Description: "Execute a shell command in the Kubernetes skill sidecar container. " +
+			Description: "Execute a shell command in a Kubernetes skill sidecar container. " +
 				"Use this to run kubectl, bash scripts, curl, jq, and other CLI tools. " +
 				"Commands execute in /workspace by default. " +
-				"If specialized MCP tools are available for the task, prefer those instead.",
+				"If specialized MCP tools are available for the task, prefer those instead. " +
+				"When multiple skill sidecars are attached, set 'target' to the name of the skill " +
+				"that owns the tool you need (e.g. 'github-gitops' for `gh`, 'k8s-ops' for `kubectl`). " +
+				"If 'target' is omitted, any sidecar may serve the request (legacy behavior).",
 			Parameters: map[string]any{
 				"type": "object",
 				"properties": map[string]any{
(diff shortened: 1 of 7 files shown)


AlexsJones (approved):
Clean, well-thought-out fix for a real problem. The approach is backwards-compatible, the test coverage is thorough, and the PR description clearly explains the race condition.

A few notes:

## Minor

1. **Duplicated shell block across 3 tool-executors** — the 13-line target routing snippet is copy-pasted identically into `skill-github-gitops`, `skill-k8s-ops`, and `skill-llmfit`. Worth extracting into a shared sourced file (e.g. `images/common/target-filter.sh`) in a follow-up to avoid drift. Not a blocker.

2. **`SYMPOZIUM_SKILL_TARGETS` is injected but never consumed** — the controller injects it into the agent container, and the PR description says the agent-runner can use it to "advise the LLM and validate" target values, but the agent-runner code in this PR doesn't read it. The tool description tells the LLM about targets via static text, but doesn't dynamically list the available targets. Fine for part 1, but the LLM has no way to know what targets are actually attached without it.

3. **Whitespace normalization mismatch** — `normalizeSidecarTarget` in Go uses `strings.TrimSpace` (trims leading/trailing only), while `tool-executor.sh` uses `tr -d '[:space:]'` (strips *all* whitespace including internal). A target like `"k8s ops"` would normalize to `"k8s ops"` in Go but `"k8sops"` in bash. Unlikely to matter in practice (SkillPack names shouldn't contain spaces), but worth aligning for correctness.

4. **No timeout on unclaimed targeted requests** — if a targeted request names a sidecar that isn't attached (typo, stale name), the request file sits in `/ipc/tools/` forever with no sidecar claiming it. The agent-runner's existing timeout covers this, but might be worth a log line in tool-executor when a request is skipped due to target mismatch.

Approving — this is a solid additive fix. The shell duplication and dynamic target list are fine as follow-ups.

PR #161, opened 2026-05-06
Enhancement: additional channel settings (part 1)

SamuZad:
Relates to https://github.com/sympozium-ai/sympozium/issues/149

This PR adds two universal channel features and one Slack-specific UX touch:

1. Stop / start keywords (universal) — keyword-driven mute / unmute per chat, persisted in a per-Agent ConfigMap so state survives controller restarts.
2. Slack emoji reactions (Slack-only) with sensible defaults
    - with the possibility of adding the other channel types (I just can't easily test those)


AlexsJones (commented):
Good PR — clean decomposition, solid test coverage, and the `callSlackAPI` consolidation is a nice improvement over the old fire-and-forget `sendMessage`. A few things to consider:

## Issues

1. **No RBAC for the channel-state ConfigMap** — the controller needs `create`/`get`/`update` on ConfigMaps in the agent's namespace for the mute store. I don't see ClusterRole changes in this diff — is this already covered by existing rules?

2. **Missing owner reference on ConfigMap** — `channel_triggers.go` creates the ConfigMap without an `ownerReferences` pointing to the Agent. If the Agent is deleted, the ConfigMap orphans. Adding an owner ref would let K8s GC clean it up automatically.

3. **ConfigMap key collisions** — the mute key format `slack:C123` could collide if a chatID ever contains `:`. Unlikely for Slack, but Discord snowflakes or other future channel types could surprise you. A safer separator (e.g. `/` or a hash) would be more defensive.

4. **Dead code in tests**:
   - `channel_reactions_test.go:805`: `recordingEventBus.mu bool` is declared but never used — remove it
   - `channel_triggers_test.go:1550`: `var _ client.Client = ...` "unused import guard" is dead code since `fake` is already used by `TestMuteStore_RoundTrip`

## Minor / non-blocking

5. **Reaction emitted before AgentRun creation** — `applyTriggers` emits `:eyes:` then returns `true`, after which the router creates the AgentRun. If AgentRun creation fails, the user sees the reaction but gets no response. Acceptable as best-effort, but worth a comment in the code.

6. **ConfigMap won't scale long-term** — one key per (channel, chatID) pair is fine for part 1, but if an agent serves hundreds of channels this gets unwieldy. Worth tracking as a follow-up for the SQLite-backed memory store (issue #45).

SamuZad:
Hi @AlexsJones, thank you for the review

# Issues
1. the necessary permissions are already declared [here](https://github.com/sympozium-ai/sympozium/blob/main/internal/controller/agentrun_controller.go#L187) and can be seen generated [here](https://github.com/sympozium-ai/sympozium/blob/main/charts/sympozium/templates/rbac.yaml#L63-L66) - so we should be good
2. Addressed
3. Addressed (via hashing)
4. Addressed

# Minor / non-blocking

5. Yes, in the "next list of channel improvements", I will add feedback for when `AgentRun` creation fails
6. Noted

AlexsJones (changes requested):
Hey @SamuZad — nice work on the channel settings feature! The trigger state machine and reaction abstraction are clean. A few things to address before merging:

**1. Rebase onto `main`**
The branch includes a bunch of already-merged commits (release 0.10.19 bump, docs pivot, stimulus node support, canvas-primitives refactor, etc.), inflating the diff from ~600 to ~2700 lines. Please rebase onto current `main` so the PR only shows the channel settings work.

**2. Handle `AlreadyExists` race in `muteStore.SetMuted`**
Two concurrent messages hitting the same chat can both read "not muted", both match a stop keyword, and both try to `Create` the ConfigMap. The second will fail with `AlreadyExists`. The create path should fall through to the update path on conflict (or use `CreateOrUpdate`/server-side apply).

**3. Use `bytes.NewReader` in `callSlackAPI`**
`strings.NewReader(string(body))` copies the `[]byte` into a string unnecessarily. `bytes.NewReader(body)` avoids the allocation.

SamuZad:
@AlexsJones thank you!
1. rebased
2. handled
3. changed

SamuZad:
@AlexsJones what do you think of the PR?

PR #167, opened 2026-05-07
Enhancement: add imagepullpolicy to various components

SamuZad:
In a lot of cases (especially with skillpack sidecar images), one might want to adjust `ImagePullPolicy` of certain images. This PR makes `ImagePullPolicy` across a number of sensible resources

The PR intentionally does not change "internal components" for now - though they can be done in a follow-up PR

### DIFF
--- a/api/v1alpha1/agent_types.go
+++ b/api/v1alpha1/agent_types.go
@@ -317,6 +317,13 @@ type SandboxSpec struct {
 	// +optional
 	Image string `json:"image,omitempty"`
 
+	// ImagePullPolicy overrides the sandbox container image pull policy.
+	// Valid values are "Always", "IfNotPresent", or "Never". Defaults to
+	// "IfNotPresent" when unset.
+	// +kubebuilder:validation:Enum=Always;IfNotPresent;Never
+	// +optional
+	ImagePullPolicy corev1.PullPolicy `json:"imagePullPolicy,omitempty"`
+
 	// Resources for the sandbox container.
 	// +optional
 	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
--- a/api/v1alpha1/agentrun_types.go
+++ b/api/v1alpha1/agentrun_types.go
@@ -166,6 +166,13 @@ type AgentRunSandboxSpec struct {
 	// +optional
 	Image string `json:"image,omitempty"`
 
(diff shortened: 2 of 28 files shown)


AlexsJones (commented):
## review

clean, well-structured PR. the centralized `ResolveImagePullPolicy()` helper with safe fallback is a good pattern, and the kubebuilder enum validation on all new fields is the right call. backward-compatible since zero value falls back to existing default. tests cover all paths. CRDs regenerated consistently across all three copies.

## issues

### 1. ensemble controller not updated (bug)

the ensemble CRD YAML was updated to include `imagePullPolicy` on `LifecycleHookContainer`, but `ensemble_controller.go:1092` still hardcodes `corev1.PullIfNotPresent` for the `memory-server` container. if ensembles can have lifecycle hooks, those hook containers in the ensemble controller should also use `ResolveImagePullPolicy()`. the CRD change without the controller change means users can set the field but it will be silently ignored.

### 2. several remaining hardcoded `PullIfNotPresent` in agentrun_controller

multiple container builds in `agentrun_controller.go` still use `corev1.PullIfNotPresent` directly (lines ~1775, 1818, 1919, 2050, 2078, 2198, 2281, 2451, 4061, 4143). most of these appear to be internal platform images (e.g. `agent-runner`, init containers), so they may be intentionally left unchanged. but this should be explicit. a brief comment in the PR description clarifying that only user-supplied images get the override would help reviewers confirm nothing was accidentally missed.

### 3. `agent_controller.go` and `model_controller.go` have remaining hardcoded instances

- `agent_controller.go:272,645` - still hardcoded.
- `model_controller.go:182` - still hardcoded.

these should be checked to determine if they're user-configurable images or platform images.

SamuZad:
@AlexsJones thanks for the review

**1. Ensemble controller / lifecycle hooks** — Not a bug. The ensemble controller doesn't build hook containers; it copies `persona.Lifecycle` straight onto stamped `Agent` CRs, and `agentrun_controller.go` ultimately builds them using `ResolveImagePullPolicy(hook.ImagePullPolicy)` for both preRun and postRun hooks. The hardcoded `PullIfNotPresent` is for the `memory-server`

**2. Remaining hardcodes in `agentrun_controller.go`** — These are intentional for now. I have updated the PR to explicitly state this

**3. agent_controller.go and model_controller.go hardcodes** — Likewise, intentional

PR #169, opened 2026-05-07
fix: propagate Memory.SystemPrompt to AgentRuns

SamuZad:
Channel router, schedule controller, and sequential successor builder
were dropping `inst.Spec.Memory.SystemPrompt` when constructing `AgentRuns`,
so steering instructions defined on the `Agent` never reached the runner.

Set `AgentRunSpec.SystemPrompt` from the parent Agent in:
- internal/controller/channel_router.go
- internal/controller/sympoziumschedule_controller.go
- internal/controller/agentrun_controller.go (sequential successor)


AlexsJones:
damn, nice spot

PR #198, opened 2026-05-15
Fixes: mcpserver volumes and arg porpagation

SamuZad:
This PR accomplishes the following:
1. adds volumes to `MCPServers`, which makes it so that they can use CSI volumes as well
2. Makes it so that args are passed, even if there is no command given
    - this is useful if you want to add additional args for an mcp server, but don't want to overwrite the default command


AlexsJones:
Nice additions! One thing worth considering:

**No validation against the reserved `adapter-bin` volume name**

The godoc on `Volumes` correctly warns users not to reuse `adapter-bin`, but nothing enforces it. If someone adds a volume with that name, the pod ends up with duplicate volume names and behavior is undefined. A webhook validation or a runtime check in `buildStdioPodSpec` that rejects/skips volumes named `adapter-bin` would make this more robust.

SamuZad:
> Nice additions! One thing worth considering:
> 
> **No validation against the reserved `adapter-bin` volume name**
> 
> The godoc on `Volumes` correctly warns users not to reuse `adapter-bin`, but nothing enforces it. If someone adds a volume with that name, the pod ends up with duplicate volume names and behavior is undefined. A webhook validation or a runtime check in `buildStdioPodSpec` that rejects/skips volumes named `adapter-bin` would make this more robust.

@AlexsJones thanks for the feedback, I added in explicit validation

PR #205, opened 2026-05-18
Introduce slack threaded responses and different and response modes

SamuZad:
This PR is related to https://github.com/sympozium-ai/sympozium/issues/149 and continues the work that began in https://github.com/sympozium-ai/sympozium/pull/161. The headliner feature is that "Slack now supports threaded responses", to reduce clutter/noise in a channel with lots of people. The more comprehensive list of changes is:
1. Slack threading support
    - agents now have the ability to respond in threads as well
    - various quality of life changes that come with multiple people talking in a channel/thread with an agent
2. Ensembles now correctly propagate `SlackOptions` to agents
3. Reworked triggering - supporting DMs and "direct @s"
4. Due to the more complex interactions, the slack channel image now logs messages and "decisions" on when/why it does or does not trigger

(I was thinking about breaking it up into chunks but ultimately they don't make sense in isolation)


AlexsJones (commented):
nice work on this one, the threading state machine is really well thought out and the test coverage is solid. a few things i'd flag:

**duplicated gating block in main.go** - the evaluateInbound + logging block is copy-pasted between handleSocketEvent and handleSlackEvents, same for the threadID promotion. if a third event path shows up this will drift. might be worth pulling into a small helper that returns the ready message and a proceed bool

**eviction on every get/update in threadEngagement** - evictLocked does a full map scan on every inbound message. fine for now but in a busy channel thats O(n) per message across all threads over 24h. a periodic sweep on a timer (every few minutes) would be cheeper under load and avoids holding the lock during eviction

**no validation on AllowedTriggers values** - the CRD accepts any string but only "mention", "dm", and "channel" actually do anything. a typo like "mentions" silently blocks everything with no feedback. worth either adding an enum constraint in the openapi schema or logging a warning in loadSlackConfig when unknown values are found

**resolveBotUserID failure makes mention-only bots unresponsive** - when auth.test fails, BotID stays empty so classifyKind can never return kindMention. if someone has allowedTriggers: ["mention"] configured, the bot goes completely silent with no obvious signal. worth either retrying with backoff or marking the pod unhealthy so it gets restarted

Happy to re-review later this pm! cheers

SamuZad:
@AlexsJones as always, thank you for the feedback!

I did my best fix the issues you highlighted

AlexsJones (approved):
all four concerns addressed nicely in 033a4a51:

- gateAndBuildInbound helper deduplicates the gating+logging+message-building across both socket mode and events api paths, clean
- eviction moved to a background sweep goroutine on a 5min ticker, get/update no longer scan the map
- loadSlackConfig now warns on unknown trigger values at startup
- resolveBotUserIDWithRetry with exponential backoff, and the fatal exit when mention-only config can't resolve the bot id is the right call, no silent broken state

looks good to merge

PR #208, opened 2026-05-18
fix(channels): propagate thread id from inbound to outbound

jevy:
## Summary

When an inbound channel message arrives in a thread (e.g. a Slack reply with `thread_ts` set), `InboundMessage` already carries the `ThreadID`, and the Slack channel pod's `sendMessage` already wires `thread_ts` when `OutboundMessage.ThreadID` is non-empty (`channels/slack/main.go:515`).

But the controller's channel router was dropping the ThreadID on the floor:

- the `AgentRun` created from the inbound message didn't carry a `reply-thread-id` annotation, and
- the `OutboundMessage` built on agent completion left `ThreadID` empty,

so `chat.postMessage` posted the reply to the channel top-level instead of threading under the user's message.

This patch stamps a `sympozium.ai/reply-thread-id` annotation on the AgentRun (when the inbound message was threaded) and reads it back when routing the response. Non-threaded inbound messages omit the annotation, so non-threaded conversations stay non-threaded — the channel pod only adds `thread_ts` when `ThreadID` is non-empty.

## Repro

1. Configure an Agent with a Slack channel (`type: slack` under `spec.channels`).
2. Bot posts a top-level message into the channel.
3. User replies in the thread under the bot's post.
4. Channel router creates an inbound AgentRun, agent completes, response posts to the channel as a **new top-level message** instead of threading.

After this patch, the reply lands in-thread under the user's message.

## CONTRIBUTING checklist

Per [CONTRIBUTING.md](CONTRIBUTING.md):

- [x] **Conventional commit scope**: `fix(controller):` (the change is in `internal/controller/`).
- [x] **`go vet`** clean.
- [x] **`go test -race -short ./internal/controller/`** passes, including two new tests:
  - `TestChannelRunAnnotations` — threaded message stamps `reply-thread-id`; non-threaded omits it.
  - `TestOutboundFromRunAnnotations` — `ThreadID` propagates when set; empty when not; missing `reply-channel` signals "not channel-sourced" so the router skips the run.
- [x] **`gofmt`** clean.
- [x] **`go build ./...`** clean.
- [x] **One concern per PR**: only the thread-id propagation.
- [x] **Docs**: behavior change is described in this PR body; no public docs need updating.

## Implementation notes

The annotation read/write logic was previously inline in two large reconciler methods. To make the threading invariant testable without spinning up a fake reconciler, I extracted two small pure helpers in `channel_router.go`:

- `channelRunAnnotations(msg *InboundMessage) map[string]string` — builds the `sympozium.ai/reply-*` annotation set. Omits `reply-thread-id` when `msg.ThreadID == ""` to keep non-threaded paths unchanged.
- `outboundFromRunAnnotations(annotations, text) (OutboundMessage, bool)` — rebuilds the OutboundMessage on completion. Returns `(_, false)` when `reply-channel` is missing so the caller skips non-channel-sourced runs (same gate as before).

## Manual verification

Tested against a homelab Sympozium v0.10.34 deployment with this patch applied (via `make run-controller` against the in-cluster NATS):

1. Bot posts a top-level "your diary is empty" message in `#agent-friend`.
2. User replies in the thread: "Can you add to memory that I have a real issue with salt..."
3. Channel router creates AgentRun with `sympozium.ai/reply-thread-id: 1779146795.121549` ✅
4. Agent completes; "Routed agent response to channel" fires ✅
5. Reply lands **in-thread** under the user's message, not top-level ✅

### DIFF
--- a/internal/controller/channel_router.go
+++ b/internal/controller/channel_router.go
@@ -271,12 +271,7 @@ func (cr *ChannelRouter) handleInbound(ctx context.Context, event *eventbus.Even
 				"sympozium.ai/source":         "channel",
 				"sympozium.ai/source-channel": msg.Channel,
 			},
-			Annotations: map[string]string{
-				"sympozium.ai/reply-channel": msg.Channel,
-				"sympozium.ai/reply-chat-id": msg.ChatID,
-				"sympozium.ai/sender-name":   msg.SenderName,
-				"sympozium.ai/sender-id":     msg.SenderID,
-			},
+			Annotations: channelRunAnnotations(&msg),
 		},
 		Spec: sympoziumv1alpha1.AgentRunSpec{
 			AgentRef:   msg.InstanceName,
@@ -383,13 +378,6 @@ func (cr *ChannelRouter) handleCompleted(ctx context.Context, event *eventbus.Ev
 		return
 	}
 
-	replyChannel := run.Annotations["sympozium.ai/reply-channel"]
-	replyChatID := run.Annotations["sympozium.ai/reply-chat-id"]
-
-	if replyChannel == "" {
-		return
-	}
-
 	// Extract the response from the completed event.
 	var result agentResult
 	if err := json.Unmarshal(event.Data, &result); err != nil {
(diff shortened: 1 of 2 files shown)


jevy:
Opened this without seeing #205 which looks more feature complete. Looking forward to that one!

PR #211, opened 2026-06-03
Allow setting additional env vars for the agent-runner (useful for max tool iterations)

SamuZad:
Adds an `env` field to `AgentConfig` and `Ensemble.spec.agentConfigs[].env` so cluster operators can inject runtime environment variables — e.g. MAX_TOOL_ITERATIONS


three-foxes-in-a-trenchcoat:
Mechanism looks solid — `AgentRunSpec.Env` already renders onto the agent container and `MAX_TOOL_ITERATIONS` is consumed in `cmd/agent-runner/main.go`, so the feature works end-to-end. DeepCopy, CRD manifests, and the Ensemble→Agent propagation are all consistent.

One gap: `Env` is threaded through 7 of the 10 `AgentRunSpec{}` construction sites, but three were missed — each already copies sibling fields like `Lifecycle`/`BaseURL` from `inst.Spec.Agents.Default`:

- `internal/webproxy/openai.go:192` (OpenAI proxy endpoint)
- `internal/webproxy/mcp.go:240` (MCP proxy endpoint)
- `internal/orchestrator/spawner.go:145` (sub-agent spawning — already does a `Get` purely to copy `Lifecycle`)

The first two matter most: the PR description claims `web-endpoint` coverage, but an agent with `env: {MAX_TOOL_ITERATIONS: ...}` would silently *not* get it when invoked via the OpenAI/MCP proxy. The spawner is more of a judgment call. Fixes are one-liners (`Env: inst.Spec.Agents.Default.Env`).

Minor: the `AgentConfig.Env` doc comment says "Values set directly on AgentRunSpec.Env take precedence" — there's no merge logic anywhere, just plain assignment from the default, so this only holds for manually-applied AgentRun CRs. Slightly misleading as written.

SamuZad:
@three-foxes-in-a-trenchcoat good catch - corrected

PR #226, opened 2026-06-18
feat: add native sidecar tool support

KSiig:
## Summary

- New `sidecar_tools.go`: loads `sidecar-tools-*.json` manifests from `/ipc/tools/`, registers them as native function-calling tools with the LLM
- Dispatch in `executeToolCall()` routes native sidecar tools through `executeSidecarTool()`, which constructs the shell command internally and uses existing IPC
- System prompt injection tells models to prefer native tools over `execute_command` for sidecar operations
- Supports `stdin` mode (pipe JSON), `args` mode (positional arguments), and `positionalArgs` for tools that need both

## Context

Part of VEL-933 Fix 7 (native sidecar tools). Companion to velatir/velatir-service-discovery#19 which adds the sidecar-side tool manifests.

With this change, models call `sd_check_service_evaluate_changes({serviceIdentifier: "...", catalog: {...}})` instead of `execute_command(target="sd-check-service", command="echo '{...}' | node /app/dist/cli.js evaluate-changes")`. This eliminates shell-quoting overhead that was the primary barrier for smaller models.

## Test plan

- [ ] `go build ./cmd/agent-runner/` passes
- [ ] `go vet ./cmd/agent-runner/` passes
- [ ] Deploy to dev with sidecar manifest changes and run enrichment agent
- [ ] Verify native tools appear in tool list and dispatch correctly
- [ ] Verify existing MCP tools and built-in tools still work

### DIFF
--- a/cmd/agent-runner/main.go
+++ b/cmd/agent-runner/main.go
@@ -260,6 +260,24 @@ func main() {
 
 			systemPrompt += sb.String()
 		}
+
+		// Load sidecar native tools from manifests
+		if sidecarTools := loadSidecarTools("/ipc/tools"); len(sidecarTools) > 0 {
+			tools = append(tools, sidecarTools...)
+
+			var sb strings.Builder
+			sb.WriteString("\n\n## Native Sidecar Tools\n\n")
+			sb.WriteString(fmt.Sprintf("You have access to %d native sidecar tools. ", len(sidecarTools)))
+			sb.WriteString("ALWAYS prefer native sidecar tools over execute_command for sidecar operations. ")
+			sb.WriteString("These tools accept structured JSON — do NOT construct shell commands.\n\n")
+			sb.WriteString("Available sidecar tools:\n")
+			for _, t := range sidecarTools {
+				sb.WriteString(fmt.Sprintf("- %s: %s\n", t.Name, t.Description))
+			}
+			systemPrompt += sb.String()
+
+			log.Printf("sidecar tools: %d tool(s) registered", len(sidecarTools))
+		}
 		log.Printf("tools enabled: %d tool(s) registered", len(tools))
 	}
 
--- a/cmd/agent-runner/sidecar_tools.go
+++ b/cmd/agent-runner/sidecar_tools.go
@@ -21,9 +21,9 @@ type sidecarToolEntry struct {
 	Description    string         `json:"description"`
 	Target         string         `json:"target"`
 	Subcommand     string         `json:"subcommand"`
-	Exec           string         `json:"exec,omitempty"`  // executable command prefix (default: "node /app/dist/cli.js")
-	InputMode      string         `json:"inputMode"`       // "stdin" or "args"
-	PositionalArgs []string       `json:"positionalArgs"`  // parameter names to pass as positional CLI args (in order)
+	Exec           string         `json:"exec,omitempty"`    // executable command prefix (default: "node /app/dist/cli.js")
+	InputMode      string         `json:"inputMode"`         // "stdin" or "args"
+	PositionalArgs []string       `json:"positionalArgs"`    // parameter names to pass as positional CLI args (in order)
 	Timeout        int            `json:"timeout,omitempty"` // per-tool timeout in seconds (0 = use default, max 120s)
 	Parameters     map[string]any `json:"parameters"`
 }
--- a/cmd/agent-runner/sidecar_tools.go
+++ b/cmd/agent-runner/sidecar_tools.go
@@ -21,10 +21,14 @@ type sidecarToolEntry struct {
 	Description    string         `json:"description"`
 	Target         string         `json:"target"`
 	Subcommand     string         `json:"subcommand"`
-	InputMode      string         `json:"inputMode"`      // "stdin" or "args"
-	PositionalArgs []string       `json:"positionalArgs"` // parameter names to pass as positional CLI args (in order)
+	Exec           string         `json:"exec,omitempty"`  // executable command prefix (default: "node /app/dist/cli.js")
+	InputMode      string         `json:"inputMode"`       // "stdin" or "args"
+	PositionalArgs []string       `json:"positionalArgs"`  // parameter names to pass as positional CLI args (in order)
+	Timeout        int            `json:"timeout,omitempty"` // per-tool timeout in seconds (0 = use default, max 120s)
 	Parameters     map[string]any `json:"parameters"`
 }
+
+const defaultSidecarExec = "node /app/dist/cli.js"
 
 var (
 	sidecarToolRegistry   = map[string]sidecarToolEntry{}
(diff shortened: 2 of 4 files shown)


AlexsJones (commented):
I actually updated this inline comments ty

KSiig:
@AlexsJones Thanks for the good and detailed feedback. I've updated the PR to hopefully align with your comments.

AlexsJones:
@KSiig - something that I've been thinking about bit more deeply is that this doesn't fit the project's security ethos: it lets a manifest in the agent-writable `/ipc` dir construct shell commands with no `SympoziumPolicy.exec` allowlist check, kpolicy that's supposed to live outside the agent's reach now effectively lives inside it.

What do you think?

KSiig:
Hmm I do think that's a fair concern. At the same time I think the functionality is necessary.

I've been running with it for the past week (custom build from a fork), and it's been game-changing. Agents are running more efficiently and more stable. Small models like Haiku was struggling to be consistent, and models from any other provider than Anthropic failed constructing the tool-calling 95% of the time.

I'll give it some thought on whether it can be implemented in another way that integrates more deeply with the security ethos, and get back to you.

AlexsJones:
Thanks for running with it on your fork and sharing the numbers, that Haiku and non Anthropic reliability jump is exactly the kind of result that tells me the feature is necessary, not optional. So I am not looking to drop it, I want to land it in a way that keeps our security and AI safety bar.

Here is the one thing I would want us to agree on first: a native sidecar tool should never grant the agent more power than `execute_command` already does. If we hold that line, native tools are just a better typed front end over the same gated path, and the posture is unchanged.

The cleanest way to get there is what I would call the full CRD path. Instead of the sidecar dropping `sidecar-tools-*.json` into the agent writable `/ipc` dir, the tools become a `tools` field on the SkillPack sidecar spec. The controller serializes that into a read only ConfigMap and mounts it on the agent, exactly the way `mcp-config` already works via `ensureMCPConfigMap`. Net effect:

- The agent consumes tool definitions but can no longer forge or rewrite them, they live in operator controlled config.
- The binary stops being the hardcoded `node /app/dist/cli.js` and becomes a declared `exec` argv on the SkillPack, so it is admission checked rather than agent supplied.
- Everything downstream you already wrote stays, the dispatch still goes `executeSidecarTool` to `executeCommand` to the exec request IPC with the same `Target` routing. One extra hardening I would fold in is passing args via the existing `execRequest.Args` list instead of building a shell string, which removes the quoting and injection surface for free.

It keeps your whole UX win, the model still sees clean function calling tools, it just changes where the definitions come from.

If you are on board with the invariant and this shape, I will build it today against `main` and we can run it head to head with your fork to make sure we keep the reliability gains. If the numbers hold, we merge that and you can retire the fork. Sound good?

KSiig:
Sounds like a great plan. The only less-than-ideal part, is that it forces users to maintain parity between the tools defined in the SkillPack, and the tools available in the sidecar. An obvious place for configuration drift to appear.

But, I think your proposal is a good way to get this implemented and un-block non-Anthropic models for now. Then we can make an explicit note in the "Writing Custom Sidecars" section of the Docs mentioning this. When describing how to register tools in the SkillPack, something like this can be added as a highlighted note:

```
NOTE: At this time, the security ethos of the project combined with the implementation of the agentRunner, prevents the agent from automatically registering tools from sidecars. This does open up the opportunity for configuration drift between sidecar tools and SkillPack definitions, but a solution to prevent this is being worked on.

Please see [this issue](https://github.com/sympozium-ai/sympozium/pull/226) for more information.
```

Then we can revisit this at a later point, but at least get the functionality out with the next release. An idea of the top of my head, would be to have a section in the docs, showing examples of scripts to include in CI/CD pipelines that run at the end of people's sidecar builds, to either patch SkillPacks manifests with `kubectl` directly, or to update Infrastructure-as-Code files.

Please let me know once you've gotten this built, then I'll go ahead and test it with my sidecars.

AlexsJones:
Opened #241 with the full CRD path we discussed above, built on top of your work here. Same UX win (native function-calling tools so Haiku and non Anthropic models stop fighting shell quoting), but the definitions move onto the SkillPack CRD and get serialized by the controller into a read only, immutable, owner referenced ConfigMap, so the agent consumes them but can never forge them. Dispatch still flows through the same gated exec IPC targeting the sidecar, now in an argv mode with no shell.

Not closing this one. I would like your eyes on #241 to make sure it keeps the reliability gains you saw on your fork, and we can decide together whether to land #241 and retire this, or keep both. I ran it end to end on a kind cluster against a local Qwen model and the native tool path worked cleanly, so I am hopeful the numbers hold.

AlexsJones:
#241 is merged to `main` — the full CRD path we landed on here. Quick recap for anyone following:

- Tool definitions live on the **SkillPack** (`sidecar.tools[]`), serialized by the controller into a read-only, immutable, owner-referenced ConfigMap mounted into the agent — so the agent consumes them but can never forge them.
- Dispatch flows through the **same gated exec IPC** as `execute_command`, now in an **argv mode** (no shell), so argument values can't inject shell syntax. The executable is the operator-declared `exec`, never model-supplied.
- Admission-validated at both AgentRun and SkillPack (snake_case names, no built-in/memory/MCP collisions, cross-pack uniqueness, positionalArgs declared + required).

**Docs are updated** so skillpack authors actually discover this:
- `concepts/skills.md` → new **Native Sidecar Tools** section (the why, the what, the trust model).
- `guides/writing-skills.md` → **Step 3b: Expose Native Tools** with a worked `sidecar.tools[]` example, plus the feature now appears in the opening "three layers" table and the quick-reference YAML.

@KSiig — I also folded in your **configuration-drift** point directly. There's now a highlighted note in the Writing Skills guide spelling out that, because definitions are operator-authored (not auto-registered from the sidecar), the author owns parity between `sidecar.tools[]` and what the sidecar implements — with the recommendation to patch the SkillPack from a CI/CD step at the end of the sidecar build (`kubectl patch` or IaC) so the two never diverge. It links back here as the tracking thread for automated reconciliation.

Whenever you get a chance to run #241 against your sidecars, would love to confirm the reliability gains hold versus your fork. If the numbers look good we can retire this PR in favor of #241; happy to keep it open until then as the drift-reconciliation tracker.

KSiig:
Tested this end-to-end with our production setup (7 SkillPacks, 28 native tools across 5 sidecar images). Confirmed it works great — all native tools registered and dispatched correctly through argv-mode IPC, including stdin-mode and positional-arg tools.

One bug I ran into during integration:

Namespace propagation bug — resolveSkillSidecars reads SkillPacks from the agent's run namespace first, where the controller's content mirror exists but without sidecar.tools[]. This means ensureSidecarToolsConfigMap is never called and the agent gets no tools manifest. I've temporarily worked around it with a direct kubectl patch but it needs a proper fix: https://github.com/sympozium-ai/sympozium/issues/249

This doesn't block adoption — the workaround is straightforward. Happy to close this PR in favour of #241 which landed the implementation.

AlexsJones:
❤️

PR #227, opened 2026-06-18
fix: remove finish_reason gate for tool call extraction

KSiig:
## Summary

- Removes the `finish_reason == "tool_calls"` check in `provider_openai.go` that silently dropped valid tool calls from OpenRouter-proxied models
- Models accessed through OpenRouter (Qwen, Gemini, Mistral) often return `finish_reason: "stop"` even when the response contains valid tool calls — this gate was discarding them
- Now checks tool call presence (`len(choice.Message.ToolCalls) > 0`) regardless of finish_reason

## Context

Part of VEL-933 tool-calling research. Benchmarks showed Qwen3-32B and Mistral models had their tool calls silently dropped by this gate, producing 0 changes despite attempting tool calls.

## Test plan

- [ ] Deploy to dev and run enrichment agent with an OpenRouter model (e.g. Qwen3-32B)
- [ ] Verify tool calls are no longer silently dropped
- [ ] Verify Anthropic and direct OpenAI models still work (they always returned `finish_reason: "tool_calls"`)

### DIFF
--- a/cmd/agent-runner/provider_openai.go
+++ b/cmd/agent-runner/provider_openai.go
@@ -150,11 +150,10 @@ func (p *openaiProvider) Chat(ctx context.Context) (ChatResult, error) {
 		FinishReason: choice.FinishReason,
 	}
 
-	// Only treat tool calls as actionable when the model signalled it's
-	// awaiting tool results. Some providers stuff tool_calls into a final
-	// message with finish_reason="stop"; the OpenAI contract is to loop
-	// only when finish_reason="tool_calls".
-	if choice.FinishReason == "tool_calls" && len(choice.Message.ToolCalls) > 0 {
+	// Extract tool calls whenever present, regardless of finish_reason.
+	// OpenRouter-proxied models (Qwen, Gemini, Mistral) often return
+	// finish_reason="stop" with valid tool calls in the same message.
+	if len(choice.Message.ToolCalls) > 0 {
 		for _, tc := range choice.Message.ToolCalls {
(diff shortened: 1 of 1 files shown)


AlexsJones (commented):
Reviewed — the fix is correct and addresses a real bug. Extracting tool calls whenever `len(choice.Message.ToolCalls) > 0` (rather than gating on `finish_reason == "tool_calls"`) is the right call: OpenRouter-proxied models (Qwen, Mistral, Gemini) return `finish_reason: "stop"` alongside valid structured tool calls, and the old gate silently dropped them. As a bonus, structured tool calls now correctly take precedence over the Qwen-recovery branch (`else if len(extraToolCalls) > 0`) — the two stay mutually exclusive, so no double-processing.

Non-blocking notes:

- **Test coverage** — there's no test that locks in this fix. The loop tests use `mockProvider` and bypass `openaiProvider.Chat` entirely, and the helper tests don't touch `Chat`. A regression here would go unnoticed. Worth a test for a completion with `FinishReason: "stop"` + non-empty `ToolCalls` → expect `result.ToolCalls` populated (extracting the extraction logic into a helper would make this easy).
- **Cross-provider consistency (informational)** — `provider_anthropic.go` and `provider_bedrock.go` still gate on the native stop reason, which is correct since those APIs guarantee the stop reason matches tool_use presence. The gate is rightly removed only on the OpenAI-compatible path.

No blockers.

AlexsJones (approved):
Approving — the nits I left above are all non-blocking. Nice fix.

PR #228, opened 2026-06-18
feat: Add skipped as AgentRun status - reachable via IPC similar to dry runs

kristof-ringleff:
This pull request introduces the ability for a preRun hook to completely skip an agent run. In a lot of cases tokens can be saved if it is clear that the LLM task is unneeded (no new emails, no issues assigned, no reports, etc).

It works similar to DryRuns however as dry runs use environment variables and these are not settable from the init container the preRun hook runs in communication via IPC was chosen.

```
kind: Ensemble
metadata:
  name: skip-run-test
spec:
  enabled: true
  description: Minimal single-agent ensemble to verify the preRun skip patch.
  category: development
  version: "0.1.0"
  agentConfigs:
  - name: skip-probe
    systemPrompt: |
      You must reply with exactly this single token and nothing else:
      NOT_SKIPPED
      
    displayName: Skip Probe
    schedule:
      type: sweep
      cron: "*/2 * * * *"
      task: "If you can read this, the run was NOT skipped — reply NOT_SKIPPED."
    lifecycle:
      preRun:
      - name: write-skip-marker
        image: "busybox:1.36"
        command:
        - sh
        - -c
        - "mkdir -p /ipc/control && printf \"test: unconditional skip\" > /ipc/control/skip && echo \"skip marker written to /ipc/control/skip\" && exit 0"
      gateDefault: block
 ```     
 
 
image

### DIFF
--- a/api/v1alpha1/agentrun_types.go
+++ b/api/v1alpha1/agentrun_types.go
@@ -270,11 +270,15 @@ const (
 	AgentRunPhaseAwaitingDelegate AgentRunPhase = "AwaitingDelegate"
 	AgentRunPhaseSucceeded        AgentRunPhase = "Succeeded"
 	AgentRunPhaseFailed           AgentRunPhase = "Failed"
+	// AgentRunPhaseSkipped is a terminal phase for runs a preRun lifecycle
+	// hook skipped (no work to do). The agent never made an LLM call, so the
+	// run consumed no tokens; it is distinct from Succeeded and Failed.
+	AgentRunPhaseSkipped AgentRunPhase = "Skipped"
 )
 
 // AgentRunStatus defines the observed state of AgentRun.
 type AgentRunStatus struct {
-	// Phase is the current phase (Pending, Running, Succeeded, Failed).
+	// Phase is the current phase (Pending, Running, Succeeded, Failed, Skipped).
 	// +optional
 	Phase AgentRunPhase `json:"phase,omitempty"`
 
--- a/charts/sympozium-crds/templates/sympozium.ai_agentruns.yaml
+++ b/charts/sympozium-crds/templates/sympozium.ai_agentruns.yaml
@@ -2637,7 +2637,7 @@ spec:
                 type: string
               phase:
                 description: Phase is the current phase (Pending, Running, Succeeded,
-                  Failed).
+                  Failed, Skipped).
                 type: string
               podName:
                 description: PodName is the name of the pod running this agent.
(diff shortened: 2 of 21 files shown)


AlexsJones (commented):
Reviewed — the design cleanly mirrors the existing DryRun mechanism, and the core state-machine work is thorough (reconcile switch, terminal/finalizer checks, delegate aggregation count, web-proxy candidate filter, CRD `phase` is enum-free so fully backward-compatible, all 6 `extractResultFromPod`/`parseAgentResultFromLogs` call sites updated). The schedule and ensemble controllers correctly need no change since they enumerate the *active* set and treat `Skipped` as terminal by fall-through.

One thing I'd recommend fixing before merge, then a few minor follow-ups:

1. **`pruneOldRuns` leaks skipped runs** (`internal/controller/agentrun_controller.go:1112`) — it collects only `Succeeded`/`Failed` for history-limit pruning, so `Skipped` runs are never pruned and accumulate indefinitely. Since this feature exists precisely to skip frequently (queue polling, empty inboxes), this set will grow fastest of all. Suggest adding `|| run.Status.Phase == "Skipped"`.

Minor (consistency, non-blocking):

2. **TUI render paths** (`cmd/sympozium/main.go` ~7368, 7563, 7710) — three `case "Succeeded", "Completed":` switches have no `Skipped` case, so a skipped run renders dim with no feed result text. The runs/pods tables and `activeOnly` filter *were* updated, so this is just an inconsistency.
3. **`buildConversationContext`** (`main.go:2822`) — renders a `Skipped` run as `[pending]` in conversation history (cosmetic; it stores its reason in `Status.Result`).
4. **Delegate skip reason dropped** (`agentrun_controller.go:868-884`) — `allTerminal` correctly counts `Skipped`, but the result-extraction branches below only handle `Failed`/`Succeeded`, so a skipped delegate child contributes neither result nor error to the parent entry. Parent recovery still works; only the surfaced result is empty.
5. **Nits** — `internal/apiserver/server.go:3412` redundant `isFailed` now recomputed by the new switch; doc comments on `AgentResult.Status` (`cmd/agent-runner/main.go:108`, `internal/ipc/protocol.go:26`) still say `"success" or "error"` and should mention `"skipped"`.

Tests cover `skipRun` and the `parseAgentResultFromLogs` skipped cases. Gaps: no test for the agent-runner skip-marker emission, the channel-router stays-silent path, or `pruneOldRuns` with skipped runs (which would have caught #1).

kristof-ringleff:
Thanks @AlexsJones for looking at this. I have addressed the missing pruning as well as the minor issues highlighted. Added test coverage where needed.

PR #229, opened 2026-06-18
feat: secretKeyRef support for env vars

kristof-ringleff:
### Problem
`preRun` hooks are documented for "fetch context, clone repos, warm caches" into the shared `/workspace`. But a hook container has **no secure way to receive a credential**, so it cannot authenticate to any private/authenticated source — e.g. pre-fetching GitHub issues or cloning a private repo with `gh`/`git`.

This PR extends secretKeyRef support for environment variables.

After PR #228 lands this can be used to test

```
kind: Ensemble
metadata:
  name: secret-prerun-test
  namespace: default
spec:
  description: "Verify secretKeyRef env injection on a preRun hook by listing private-repo issues with a PAT."
  category: development
  version: "0.1.0"
  workflowType: delegation
  relationships:
    - source: kickoff
      target: secret-probe
      type: stimulus
  stimulus:
    name: kickoff
    prompt: "Probe run — the preRun should fetch issues with the PAT, then skip the LLM."
  agentConfigs:
    - name: secret-probe
      displayName: "Secret Probe"
      systemPrompt: |
        Reply with exactly this single token and nothing else:
        NOT_SKIPPED
      lifecycle:
        preRun:
          - name: list-issues-with-pat
            # github-gitops image ships `gh`, which auto-authenticates from $GH_TOKEN.
            image: ghcr.io/sympozium-ai/sympozium/skill-github-gitops:latest
            env:
              - name: GH_TOKEN
                valueFrom:
                  secretKeyRef:
                    name: github-gitops-token
                    key: GH_TOKEN
              - name: SKILL_REPO
                value: "example/test"
            command:
              - sh
              - -c
              - |
                mkdir -p /workspace/context /ipc/control
                if [ -z "$GH_TOKEN" ]; then
                  echo "GH_TOKEN is EMPTY — secretKeyRef not injected"
                  printf 'FAIL: GH_TOKEN empty — secretKeyRef not injected' > /ipc/control/skip
                  exit 0
                fi
                if ! COUNT=$(gh issue list --repo "$SKILL_REPO" --state open --json number --jq 'length' 2>/tmp/gh.err); then
                  echo "gh issue list FAILED:"; cat /tmp/gh.err
                  printf 'FAIL: gh issue list errored: %s' "$(cat /tmp/gh.err)" > /ipc/control/skip
                  exit 0
                fi
                gh issue list --repo "$SKILL_REPO" --state open --json number,title > /workspace/context/issues.json
                echo "OK: secretKeyRef works — $SKILL_REPO has $COUNT open issues"
                printf 'OK: secretKeyRef works — %s has %s open issues' "$SKILL_REPO" "$COUNT" > /ipc/control/skip
                exit 0
      # Exercise the SCHEDULED path (the schedule creator copies the Agent's
      # lifecycle — incl. valueFrom — into each AgentRun). 
      schedule:
        type: sweep
        cron: "*/2 * * * *"
        task: "Scheduled probe — the preRun should inject the PAT via secretKeyRef and skip the LLM."
```


AlexsJones (commented):
Reviewed — clean implementation. The controller never reads the secret (kubelet resolves `SecretKeySelector` at pod start), so there's no leakage into logs or status. I re-ran `make generate manifests` and got zero drift — deepcopy and all four CRD manifest copies are in sync, which is the most common failure mode for this class of API change and it's handled correctly here.

Non-blocking notes:

1. **Validation gap for skill sidecar env** (`internal/webhook/policy_enforcer.go:355-374`) — the new `validateHookEnv` mutual-exclusivity/completeness check is wired only into the AgentRun PreRun/PostRun path. The same `EnvVar` type is reused by `SkillSidecar.Env`, which got the `valueFrom` field but no equivalent validation, so a sidecar entry with both `value` and `valueFrom` won't be rejected (`toCoreEnvVar` silently prefers `valueFrom`). Graceful, but inconsistent — fine if sidecar specs are operator-trusted, otherwise worth mirroring the check.

2. **`secretKeyRef` not governed by `LifecyclePolicy.DeniedResources`** (`policy_enforcer.go:385`) — denying the `secrets` RBAC resource for hooks doesn't constrain `secretKeyRef` env injection; these are independent. This matches the standard k8s pod-author trust model (same as native `env.valueFrom.secretKeyRef`), so it's a reasonable choice — just flagging in case a Secret-name allowlist belongs in `LifecyclePolicySpec` later.

3. **Nit** (`api/v1alpha1/skillpack_types.go:47-51`) — doc comment says "in the AgentRun namespace" on the shared `SecretKeySelector`/`EnvVarSource` types; "in the same namespace" reads better for the sidecar case.

Tests cover the three meaningful webhook branches plus the controller conversion. No blockers.

AlexsJones (approved):
Approving — notes above are non-blocking. CRD/deepcopy regen verified clean.

PR #230, opened 2026-06-19
feat: Persist AgentRun.spec.Timeout and expose runTimeout via Ensemble config

### LINKED ISSUE #222: Scheduled AgentRuns can't be given a timeout > 10 minutes
### Environment
Sympozium **v0.10.36** (tag `7ffce2e`). `provider: custom` (self-hosted OpenAI-compatible model). Single controller replica.

### Problem
A scheduled agent whose run needs more than 10 minutes always fails with `status.error: "timeout"` at 10:00 (the pod is deleted), regardless of how the timeout is configured. There is no operator-facing way to raise it for scheduled/sweep runs.

The reason is that a run is bounded by **four independent timeout gates**, set by three different mechanisms, and none of the configurable ones reaches the gate that actually fires:

| Gate | Default | Set by | Reachable for scheduled runs? |
|---|---|---|---|
| Runner `run_timeout` (whole run) | 10m cloud / 30m local | `RUN_TIMEOUT` env | ✅ via `agentConfigs[].env` (since `ae5e1e6`) |
| Runner `llm_request_timeout` (per call) | 0 cloud / 5m local | `LLM_REQUEST_TIMEOUT` env | ✅ via `agentConfigs[].env` |
| Job `activeDeadlineSeconds` | 600s (10m) | `Agent.spec.agents.default.runTimeout` | ⚠️ only via out-of-band `Agent` patch (no Ensemble field) |
| **Controller watchdog** | **10m hard default** | persisted `AgentRun.spec.Timeout` | ❌ **never set for scheduled runs → always 10m** |

The first three can now be raised in v0.10.36, but the **controller watchdog** is the binding gate and cannot be configured, so the run dies at 10m anyway.

### Root cause
`internal/controller/agentrun_controller.go` (L727–740) — runs every reconcile while the pod is alive:

```go
timeout := 10 * time.Minute // default hard timeout
if agentRun.Spec.Timeout != nil {
    timeout = agentRun.Spec.Timeout.Duration
}
if elapsed > timeout {
    _ = r.Delete(ctx, job, ...)              // kills the pod
    return ctrl.Result{}, r.failRun(ctx, agentRun, "timeout")
}
```

The only override is the **persisted** `AgentRun.spec.Timeout`, and nothing sets it for scheduled runs:
- `sympoziumschedule_controller.go` (L234) builds the `AgentRunSpec` copying `Env` but **not** `Timeout`.
- `agentrun_controller.go` (~L389) does copy `Agent.runTimeout` into `spec.Timeout`, but only **in-memory** during Job construction — it sets the Job `activeDeadlineSeconds` (~L1749) and is never written back, so the watchdog (and the `RUN_TIMEOUT` env injection at ~L1828) read `nil`.

Net: `Agent.runTimeout` raises the Job deadline only; the runner self-terminates at the provider default and the watchdog fires at 10m.

### What would fix it
Persist `AgentRun.spec.Timeout` from `Agent.spec.agents.default.runTimeout` in the run creators (the same place `ae5e1e6` added `Env`, e.g. `sympoziumschedule_controller.go`):

```go
Timeout: parseRunTimeout(instance.Spec.Agents.Default.RunTimeout), // *metav1.Duration, nil if unset
```

A single persisted `spec.Timeout` then drives **all** the controller-side gates consistently — the watchdog (L731), the Job `activeDeadlineSeconds` (L1749), and the `RUN_TIMEOUT` env (L1828) — so one operator setting works.

Ideally, also expose `runTimeout` on `Ensemble.spec.agentConfigs[]` (like `env` already is) so the budget is fully declarative and no out-of-band `Agent` patch is needed.

### Minimal reproduction
1. Scheduled (`sweep`) agent on a model slow enough to take > 10 min per run.
2. Set `agentConfigs[].env.RUN_TIMEOUT: "30m"` and patch `Agent.runTimeout: "32m"`.
3. Trigger a sweep with real work.

Expected: ~30 min. 
Actual: `Failed` / `error: "timeout"` at exactly 10m, pod deleted. (Confirmed: Job `activeDeadlineSeconds=1980`, runner log `run_timeout=30m0s`, yet the run is killed at 10:00 — the watchdog's 10m default.)

kristof-ringleff:
Fixes #222 allowing scheduled agent runs to take longer than the hardcoded 10 minutes.


AlexsJones (commented):
Reviewed — core change is correct and the thesis is right: the controller watchdog (`agentrun_controller.go:730`) hard-caps runs at 10m unless `Spec.Timeout` is non-nil, so persisting it at creation is the only reliable lever. I re-ran `make manifests generate` and got zero drift, and the new `ParseRunTimeout` helper nicely de-duplicates the previously-inline parsing.

Non-blocking notes:

1. **Incomplete rollout in webproxy** (`internal/webproxy/openai.go:208`, `internal/webproxy/mcp.go:254`) — both still hardcode `Timeout: &metav1.Duration{Duration: 10 * time.Minute}` and weren't updated. Because they set a *non-nil* 10m timeout, the new `reconcilePending` fallback (which only fires when `Spec.Timeout == nil`) won't correct them — so web-proxy-driven runs silently ignore a configured `runTimeout`. Both have `inst` in scope and could call `ParseRunTimeout()` like the other sites. Not a blocker for the scheduled/sweep case in #222, but it contradicts the "single persisted Timeout drives all gates" goal.

2. **Doc-comment accuracy** (`api/v1alpha1/ensemble_types.go:281`, `agent_types.go:391`) — both promise a "provider-appropriate default" when empty, but that cloud-10m/local-30m default only exists in the runner (`cmd/agent-runner/main.go:98-103`); the controller watchdog uses a flat 10m when `Spec.Timeout` is nil. So a local provider with no `runTimeout` gets killed by the watchdog at 10m even though the runner would self-permit 30m. Pre-existing inconsistency, not introduced here, but the comments reinforce a narrative the controller doesn't honor.

3. **Nit** (`internal/controller/channel_router.go:264-266`) — explicitly falls back to `10 * time.Minute` while every other updated site leaves it nil and relies on the watchdog default. Functionally equivalent, just stylistically divergent.

`ParseRunTimeout` has a solid table test and the schedule persistence path is covered. No blockers.

AlexsJones (approved):
Approving — core fix is correct. The webproxy/doc-comment notes above are non-blocking follow-ups.

PR #234, opened 2026-06-24
fix: preserve namespace for subagent spawning

### LINKED ISSUE #233: spawn_subagents appears to fail when parent AgentRun is not in default namespace
Hi, I believe there may be a namespace handling issue in the `spawn_subagents` flow.

We are running Sympozium in a dedicated namespace, `sympozium-system`, and we observed that a parent `AgentRun` in that namespace was able to call the `spawn_subagents` tool, but no child `AgentRun`s were created.

From the controller logs, it looked like the spawn router received the subagent request, but then failed while trying to look up the parent `AgentRun`:

```text
failed to look up parent AgentRun for subagent request
```

I believe the issue may be that the spawn router looks for the parent `AgentRun` in the `default` namespace. In our case, the parent run exists in `sympozium-system`, so the lookup fails before child runs can be created.

The relevant code seems to be around the parent lookup in `internal/controller/spawn_router.go`, where the namespace appears to be hardcoded to `default`.

I also noticed that `AGENT_NAMESPACE` seems to be injected into agent pods, but I do not think the IPC bridge currently includes that namespace in the spawn request event metadata. So the spawn router may not have enough information to resolve the parent run namespace reliably.

Expected behavior:

- A parent `AgentRun` in `sympozium-system` should be able to call `spawn_subagents`.
- Child `AgentRun`s should be created in the same namespace as the parent.
- The spawn router should resolve the parent run using the actual parent namespace, not always `default`.

Observed behavior:

- Parent `AgentRun` in `sympozium-system` calls `spawn_subagents`.
- Spawn router receives the request.
- Parent lookup fails.
- No child `AgentRun`s are created.

I believe a possible fix would be:

- Have `ipc-bridge` read `AGENT_NAMESPACE`.
- Include the namespace in event metadata for spawn and subagent requests.
- Have `spawn_router.go` use that namespace when looking up the parent `AgentRun`.
- Keep a fallback to `default` for compatibility with older events.

We may have misunderstood how this is intended to work, but based on the behavior and logs, it looks like non-default namespaces currently break subagent spawning.

dag-andersen:
## Summary

This updates the subagent spawn path to carry the parent AgentRun namespace through IPC bridge event metadata and use it in the spawn router.

I believe this fixes #233, where `spawn_subagents` can fail when the parent AgentRun is not in the `default` namespace.

## Details

- Read `AGENT_NAMESPACE` in `cmd/ipc-bridge`
- Include `namespace` in standard IPC bridge event metadata
- Use the parent namespace when looking up the parent AgentRun in `SpawnRouter`
- Use the parent namespace for delegate status and circuit breaker lookups
- Keep `default` as a fallback when namespace metadata is missing
- Add regression coverage for non-default namespace circuit breaker lookup

## Testing

```text
go test ./internal/controller -run 'TestCircuitBreaker|TestPendingBatch'
go test ./internal/ipc ./cmd/ipc-bridge
go test ./...
```

### DIFF
--- a/cmd/ipc-bridge/main.go
+++ b/cmd/ipc-bridge/main.go
@@ -20,11 +20,13 @@ func main() {
 	var basePath string
 	var agentRunID string
 	var instanceName string
+	var agentNamespace string
 	var eventBusURL string
 
 	flag.StringVar(&basePath, "ipc-path", "/ipc", "Base path for IPC directory")
 	flag.StringVar(&agentRunID, "agent-run-id", os.Getenv("AGENT_RUN_ID"), "Agent run ID")
 	flag.StringVar(&instanceName, "instance", os.Getenv("INSTANCE_NAME"), "Agent name")
+	flag.StringVar(&agentNamespace, "namespace", os.Getenv("AGENT_NAMESPACE"), "Agent namespace")
 	flag.StringVar(&eventBusURL, "event-bus-url", os.Getenv("EVENT_BUS_URL"), "Event bus (NATS) URL")
 	flag.Parse()
 
@@ -49,7 +51,7 @@ func main() {
 	ctx, cancel := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
 	defer cancel()
 
-	bridge := ipc.NewBridge(basePath, agentRunID, instanceName, bus, log)
+	bridge := ipc.NewBridge(basePath, agentRunID, instanceName, bus, log, agentNamespace)
 	if err := bridge.Start(ctx); err != nil {
 		log.Error(err, "bridge failed")
 		os.Exit(1)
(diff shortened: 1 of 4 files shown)


AlexsJones (commented):
Thanks for this, the fix is the right shape and it is complete. The namespace now threads end to end: ipc-bridge reads `AGENT_NAMESPACE`, `metadata()` stamps it on every event, and `spawn_router` resolves the parent via `lookupParentRun` with `namespaceOrDefault` keeping the `default` fallback for older bridges. I checked all the previously hardcoded lookups (parent, child completion, subagent, the delegate status update, and both circuit breaker paths) and they are all resolved, with the batch sequential path already using `batch.namespace`. Nice and thorough.

A couple of non blocking nits:

1. `NewBridge(..., agentNamespace ...string)` is variadic but has a single caller, a plain `agentNamespace string` reads cleaner and avoids silent misuse (`internal/ipc/bridge.go:47`).
2. `NewBridge` does its own `os.Getenv("AGENT_NAMESPACE")` fallback, but the flag in `main.go` already defaults to that env, so it is dead for the real caller.
3. Would be good to add a small `internal/ipc/bridge_test.go` asserting `metadata()` includes `namespace`, so the event side of the fix has coverage. The new `TestCircuitBreaker_UsesParentNamespace` covers the router side nicely.

None of these block. Happy to merge once the nits are tidied, or I can take them in a follow up. Great catch and great fix.

PR #238, opened 2026-06-25
feat(agent-runner): add opt-in detailed JSONL file logging with per-run directories

KSiig:
## Summary

- Add opt-in detailed JSONL file logging for agent runs (`DETAILED_LOG_PATH` / `DETAILED_LOG_MAX_SIZE`)
- `agent.jsonl` and `llm.jsonl` capture complete, untruncated tool call/response pairs and LLM interactions
- Shared monotonic sequence numbering across all 12 logging sites
- When the run ID is known, logs are written to a per-run subdirectory (`<path>/<run-name>/`) so each run's logs are preserved independently

## Context

Based on [KSiig/sympozium PR #6](https://github.com/KSiig/sympozium/pull/6) (merged into fork), rebased onto current main and enhanced with per-run subdirectory support. Validated through 12 automated check-in iterations against the `velatir-agents-dev` cluster.

## Test plan

- [ ] Set `DETAILED_LOG_PATH=/var/log/sympozium` and mount a writable volume
- [ ] Run an agent — verify `agent.jsonl` and `llm.jsonl` appear under `<path>/<run-name>/`
- [ ] Verify log entries have monotonically increasing sequence numbers
- [ ] Verify file rotation triggers at `DETAILED_LOG_MAX_SIZE`
- [ ] Verify no logs are written when `DETAILED_LOG_PATH` is unset

### DIFF
--- a/cmd/agent-runner/canary.go
+++ b/cmd/agent-runner/canary.go
@@ -73,6 +73,7 @@ func checkAPIServer(ctx context.Context) canaryCheck {
 	if strings.TrimSpace(body) == "ok" {
 		return canaryCheck{Name: "API Server", Status: "pass", Details: "ok"}
 	}
+	detailedLog.LogAgent("canary_api_fail", map[string]any{"body": body})
 	return canaryCheck{Name: "API Server", Status: "fail", Details: "unexpected response: " + truncate(body, 100)}
 }
 
@@ -258,6 +259,7 @@ func httpGet(ctx context.Context, url string, timeout time.Duration) (string, er
 		return "", err
 	}
 	if resp.StatusCode != http.StatusOK {
+		detailedLog.LogAgent("canary_http_error", map[string]any{"status_code": resp.StatusCode, "body": string(b)})
 		return "", fmt.Errorf("HTTP %d: %s", resp.StatusCode, truncate(string(b), 200))
 	}
 	return string(b), nil
--- a/cmd/agent-runner/detailed_logging.go
+++ b/cmd/agent-runner/detailed_logging.go
@@ -1,0 +1,189 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"log"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"time"
+)
+
+var detailedLog *DetailedLogger
+
+// DetailedLogger writes untruncated JSONL logs to agent.jsonl and llm.jsonl.
+type DetailedLogger struct {
+	agentFile *os.File
+	llmFile   *os.File
+	seq       atomic.Int64
+	maxSize   int64
+	runID     string
+	mu        sync.Mutex
+}
+
+// NewDetailedLogger opens (or creates) agent.jsonl and llm.jsonl under path.
+func NewDetailedLogger(path, runID string, maxSize int64) (*DetailedLogger, error) {
+	if err := os.MkdirAll(path, 0o755); err != nil {
+		return nil, fmt.Errorf("creating log dir: %w", err)
+	}
+
+	agentFile, err := os.OpenFile(filepath.Join(path, "agent.jsonl"), os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0o644)
+	if err != nil {
+		return nil, fmt.Errorf("opening agent.jsonl: %w", err)
+	}
+
+	llmFile, err := os.OpenFile(filepath.Join(path, "llm.jsonl"), os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0o644)
+	if err != nil {
+		agentFile.Close()
+		return nil, fmt.Errorf("opening llm.jsonl: %w", err)
+	}
+
+	return &DetailedLogger{
+		agentFile: agentFile,
+		llmFile:   llmFile,
+		maxSize:   maxSize,
+		runID:     runID,
+	}, nil
+}
+
+// LogAgent writes an untruncated event to agent.jsonl. No-op on nil receiver.
+func (dl *DetailedLogger) LogAgent(event string, fields map[string]any) {
+	if dl == nil {
+		return
+	}
+	dl.writeEntry(true, event, fields)
+}
+
+// LogLLM writes an untruncated event to llm.jsonl. No-op on nil receiver.
+func (dl *DetailedLogger) LogLLM(event string, fields map[string]any) {
+	if dl == nil {
+		return
+	}
+	dl.writeEntry(false, event, fields)
+}
+
+func (dl *DetailedLogger) writeEntry(isAgent bool, event string, fields map[string]any) {
+	seq := dl.seq.Add(1)
+	entry := map[string]any{
+		"ts":     time.Now().Format(time.RFC3339Nano),
+		"run_id": dl.runID,
+		"seq":    seq,
+		"event":  event,
+	}
+	for k, v := range fields {
+		entry[k] = v
+	}
+
+	data, err := json.Marshal(entry)
+	if err != nil {
+		log.Printf("detailed_logging: marshal error: %v", err)
+		return
+	}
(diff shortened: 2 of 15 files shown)


AlexsJones (commented):

Automated multi-angle review; every finding adversarially verified against this branch with quoted code. The feature shape is good (nil-receiver no-op, opt-in via env, nothing existing removed) — the issues are in the logger itself.

1. **Secrets persisted world-readable to durable storage.** All four `OpenFile` calls use `0o644` (dir `0o755`), and content is untruncated by design: full exec output, full file contents from `read_file`, full LLM payloads. The docs recommend a PVC "for durable storage", so credentials echoed by a command or read from a config end up on a durable, mountable volume — a broader and longer-lived surface than kubectl-logs RBAC. Suggest `0o600`/`0o700`, and a redaction/security warning in `docs/guides/detailed-logging.md`.

2. **`detailed_logging.go:175` — `DETAILED_LOG_MAX_SIZE=0m` (or any negative value) destroys the logs.** `parseSize` has no positivity check; with `maxSize <= 0`, every write triggers `rotate()`, and `os.Rename` clobbers the single `.1.jsonl` backup each time — only the most recent entry survives.

3. **`detailed_logging.go:110` — the size cap is checked only after the write, and entries are unbounded.** `read_file` on a 300MB file lands as one JSONL line regardless of the cap (logging happens before the 8,000-char truncation) — real emptyDir-exhaustion / pod-eviction risk. Consider a per-entry cap.

4. **`detailed_logging.go:74` — `seq` resets on container restart while files open `O_APPEND`.** A restart with the same `AGENT_RUN_ID` on a PVC appends a second `seq=1,2,3…` stream to the same files, silently corrupting the documented `jq -s 'sort_by(.seq)'` merge recipe. Seed from the existing file (or include a process epoch in the sort key).

5. **`mcp_tools.go:253` — failed MCP calls are never logged.** The `mcp_tool_result` log call sits after `formatMCPResult`'s early returns (`!r.Success`, `IsError`, unmarshal failure) — exactly the failure cases this feature exists to debug produce no result entry.

6. **`detailed_logging.go:143` — `rotate()` ignores the `os.Rename` error** and reopens with `O_APPEND|O_CREATE`; if the rename fails persistently, it reopens the same oversized file and grows past the cap forever, with only the (never-failing) reopen error checked.

7. **`detailed_logging.go:192` — `parseSize` silently returns the 50MB default** for plain byte counts, `k`, `kb`, or `mb` inputs — no warning anywhere, so a misconfigured cap is invisible until the volume fills. Return an error and route it through the existing "detailed logging disabled" warning in `main.go`.

8. **Coverage holes + per-provider drift (altitude).** Result logging is hand-placed inside individual tools, so `write_file`, `list_directory`, `send_channel_message`, `schedule_task`, `spawn_subagents`, memory/workflow-memory and sidecar tools emit *no* result entries — despite the docs promising a "complete chronological view". LLM logging is copy-pasted into all three providers with divergent schemas (`error_code` vs `status_code`; `content` vs `text` vs `choices`), and a fourth provider would silently ship unlogged. Logging once at the `executeToolCall` dispatcher (which already sees every tool name/args/result) and as a decorator around `Provider.Chat` fixes coverage, schema stability, and the `formatMCPResult` signature-threading in one place.

Perf notes (same hot path, fix together if it matters): a `Stat()` syscall after every write under one global mutex shared by both streams; unbuffered one-`write(2)`-per-entry; and the `map[string]any` argument is allocated at all ~22 call sites even when logging is disabled (the default) — track file size in an int64 and guard call sites with an enabled check (or accept a lazy `func() map[string]any`).

AlexsJones (approved):
Delta re-review of the four new commits (`6487ede` → `bf4b8aa`) — all eight findings from the previous review are addressed, and mostly in the structural way rather than the patch-the-symptom way:

1. **Permissions**: dir `0o700`, both files `0o600`. ✓
2. **`parseSize` positivity**: now returns `size must be positive` / `invalid size` errors instead of silently rotating everything away. ✓
3. **Unbounded entries**: 1 MB per-entry cap with explicit `(truncated by detailed logger)` marker, applied before the write. ✓
4. **`seq` restart corruption**: `seedSeqFromFiles` resumes from the existing files, so the `jq -s 'sort_by(.seq)'` recipe stays valid across container restarts on a PVC. ✓
5. **Failed MCP calls**: covered by the structural fix below — result logging now happens in `executeToolCallWithTelemetry` *after* the call returns, so `MCP Error:` results are logged with `is_error: true` instead of vanishing in `formatMCPResult`'s early returns. ✓
6. **`rotate()` rename failure**: now logs and disables the stream rather than regrowing the same oversized file forever. ✓
7. **Silent size-parse fallback**: routed through the disable-with-warning path in `main.go`. ✓
8. **Coverage holes / per-provider drift**: this is the one I'm happiest with — `tool_call`/`tool_result` logging moved to the dispatcher (`executeToolCall` / `executeToolCallWithTelemetry`), so every tool including future ones is covered, and the LLM logging is centralized behind the provider seam instead of copy-pasted per provider. ✓

The perf notes landed too: 64 KB buffered writers and size tracked in `int64` counters instead of a `Stat()` per write. The accidentally committed binary is removed and the branch is gofmt-clean; CI green.

Approving — recommending to @AlexsJones for final look and merge.

AlexsJones (changes requested):
Thanks @KSiig — this is a genuinely useful debugging feature, and the design is nice: the nil-receiver no-op pattern keeps call sites clean, the shared `seq` counter for merging both files is a good idea, and warn-and-continue is the right failure mode for an opt-in diagnostic. Docs are thorough too. A few things before merge:

1. **Add unit tests for `detailed_logging.go`.** Every other file in `cmd/agent-runner` has a paired `_test.go`; this one has none, and `parseSize`, `writeEntry`, and `rotate` are all easily testable with `t.TempDir()`. At minimum: parseSize cases (empty, `50m`, `1g`, plain number, garbage), rotation at maxSize, and nil-receiver safety.
2. **Security posture of the log files.** These intentionally capture untruncated LLM responses, full `read_file` contents, exec commands and outputs, fetched URL bodies, and the raw task — exactly where secrets and PII end up. Please (a) create the directory `0700` and files `0600` instead of `0755`/`0644`, and (b) add a prominent warning to `docs/guides/detailed-logging.md` that these logs may contain credentials/PII and should be mounted and retained accordingly.
3. **`parseSize` silently falls back on bad input.** `DETAILED_LOG_MAX_SIZE: "100"` (no suffix) or a typo like `50mb` silently becomes the 50MB default with no signal to the operator. Accept a plain byte count and/or warn when falling back.
4. **Sanitize `AGENT_RUN_ID` before `filepath.Join`.** A run ID containing separators or `..` would escape `DETAILED_LOG_PATH`; `filepath.Base(runID)` (or rejecting separators) closes this cheaply.
5. **Docs vs. behaviour mismatch for `llm.jsonl`.** The guide says "full LLM request and response payloads", but `request` events only log counts (`messages_count`, `system_len`, `tools_count`), and the Bedrock `response` logs only `textContent` while Anthropic/OpenAI log full content blocks. Make doc match code (or code match doc — consistently across providers). The doc example also shows `"seq":0` but the counter starts at 1.
6. **Multi-writer rotation race when `AGENT_RUN_ID` is empty.** With a shared PVC and no run ID, multiple runner pods append to the same flat `agent.jsonl`, and stat-then-rename rotation isn't safe across processes. Per-run subdirs already solve this when the run ID is set — a doc note that a shared flat path is unsupported for concurrent runs is enough.
7. **Nits:** `formatMCPResult` now logs as a side effect of a pure formatting function (and its early error returns skip logging) — consider logging at the call site in `executeMCPTool`; `writeEntry` calls `f.Stat()` per write where tracking bytes in the struct avoids a syscall; the new doc is missing a trailing newline.

Also — could you link the motivating issue/discussion if there is one? The PR references your fork's PR #6 but nothing in this repo.

None of this blocks the overall approach; happy to re-review quickly once tests and the permissions/docs items are in.

PR #240, opened 2026-06-25
fix(controller): propagate runTimeout and wire agentEnv into pod spec

KSiig:
## Summary
Two follow-up bugs from #230 (runTimeout via Ensemble config):

- **Ensemble controller**: `runTimeout` was propagated when *creating* new Agent instances from Ensemble agentConfigs, but missed the update-existing-instance path. When an Ensemble's `agentConfig.runTimeout` changed, the Agent kept its stale value.
- **AgentRun controller**: `agentEnv` (containing `RUN_TIMEOUT`, `TRACEPARENT`, OTEL vars) was computed but never appended to `containers[0].Env` — the pod spec used a hardcoded inline env list, so the computed vars were dead code.

## Changes
- `ensemble_controller.go`: add `RunTimeout` to the reconciliation block that already propagates Model, BaseURL, ProviderHeaders, Env, and Skills changes
- `agentrun_controller.go`: append `agentEnv` to `containers[0].Env` so computed vars reach the pod

## Test plan
- [x] Verified on dev cluster: Ensemble `runTimeout: 30m` propagates to Agent CR, AgentRun, and pod logs show `run_timeout=30m0s`
- [x] Verified `DETAILED_LOG_PATH` and `MAX_TOOL_ITERATIONS` env vars reach the pod container
- [x] Verified `RUN_TIMEOUT` env var reaches the pod container

### DIFF
--- a/internal/controller/agentrun_controller.go
+++ b/internal/controller/agentrun_controller.go
@@ -2037,6 +2037,9 @@ func (r *AgentRunReconciler) buildContainers(
 		})
 	}
 
+	// Inject computed env vars (RUN_TIMEOUT, TRACEPARENT, OTEL).
+	containers[0].Env = append(containers[0].Env, agentEnv...)
+
 	// Inject custom environment variables from AgentRun spec.
 	// Sort keys for deterministic pod specs.
 	envKeys := make([]string, 0, len(agentRun.Spec.Env))
--- a/internal/controller/agentrun_controller.go
+++ b/internal/controller/agentrun_controller.go
@@ -1870,18 +1870,9 @@ func (r *AgentRunReconciler) buildContainers(
 				},
 			},
 			Env: []corev1.EnvVar{
-				{Name: "AGENT_RUN_ID", Value: agentRun.Name},
-				{Name: "AGENT_ID", Value: agentRun.Spec.AgentID},
-				{Name: "SESSION_KEY", Value: agentRun.Spec.SessionKey},
 				{Name: "INSTANCE_NAME", Value: agentRun.Spec.AgentRef},
 				{Name: "ENSEMBLE_NAME", Value: agentRun.Labels["sympozium.ai/ensemble"]},
 				{Name: "AGENT_NAMESPACE", Value: agentRun.Namespace},
-				{Name: "TASK", Value: agentRun.Spec.Task},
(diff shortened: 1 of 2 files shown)


AlexsJones:
Confirmed this fixes a **real latent bug**, not just a tidy-up. On `main`, `agentEnv` (built in `buildContainers`, holding `AGENT_RUN_ID`, `AGENT_ID`, `RUN_TIMEOUT`, `TRACEPARENT`, `OTEL_*`) is assembled but **never attached to the agent container** — the `agent` container uses its own inline `Env: []corev1.EnvVar{...}` literal. So `RUN_TIMEOUT` (from #230) and OTel traceparent were silently dropped for the agent. Wiring `agentEnv` in is the right fix. 👍

One thing to clean up before merge, though: the inline `Env` literal already sets these 7 keys, which also live in `agentEnv`:

`SESSION_KEY`, `TASK`, `SYSTEM_PROMPT`, `MODEL_PROVIDER`, `MODEL_NAME`, `MODEL_BASE_URL`, `THINKING_MODE`

So `containers[0].Env = append(containers[0].Env, agentEnv...)` will emit **duplicate env entries** for each. Kubernetes tolerates duplicates (last value wins), but it's confusing in `kubectl describe` and a latent footgun if the two ever diverge.

Suggested fix: make `agentEnv` the single source of truth — drop the overlapping keys from the inline literal and let the append provide them (keeping container-only keys like `INSTANCE_NAME`, `ENSEMBLE_NAME`, `AGENT_NAMESPACE` in the literal). Either direction is fine as long as each key appears once. Happy to approve once that's deduped.

KSiig:
@AlexsJones Appreciate the feedback, I've implemented your suggestion.

KSiig:
Added Env update propagation in commit 48c2f77. This was the missing third piece alongside RunTimeout and Subagents — without it, the Ensemble controller never synced env map changes (like MAX_TOOL_ITERATIONS, DETAILED_LOG_PATH) to existing Agent instances. New Agents got the env on
  creation, and the AgentRun copy path worked, but updates to the Ensemble's env map were silently dropped for already-existing Agents.

  Discovered this because enrichment runs were hitting the default MAX_TOOL_ITERATIONS=50 despite the Ensemble setting it to 250 — the Agent instance predated the env config and was never updated.

PR #241, opened 2026-06-25
feat: native sidecar tools declared on SkillPack CRD

AlexsJones:
## Summary

Native, typed function-calling tools that a SkillPack sidecar exposes to the agent's LLM — the CRD-path alternative to #226. Instead of the model constructing an `execute_command` shell string, it calls a tool like `sd_evaluate_changes({serviceIdentifier: "...", catalog: {...}})`. Small/local and non-Anthropic models are dramatically more reliable with native tools than with shell-quoting.

The design holds one invariant: **a native tool grants no more authority than `execute_command` already does**, and tool definitions live in operator-controlled config the running agent cannot forge.

- Tools are declared on the **SkillPack CRD** (`sidecar.tools[]`), so they're operator-authored and admission-validated.
- The **controller** serializes them into a per-run, **read-only, immutable, owner-referenced ConfigMap** mounted into the agent. The IPC routing `target` is controller-derived from the SkillPack name, never model-supplied.
- Dispatch flows through the **same gated exec IPC** as `execute_command`, in a new **argv mode** (no shell), so argument values can't inject shell syntax. The executable comes from the operator-declared `exec`, never the model.

This mirrors the MCP precedent (controller-written config, read-only mount, struct-mirroring across the controller/agent-runner boundary).

## What's in here

- **api**: `SidecarTool` type + `SkillSidecar.tools[]` (exec argv, subcommand, inputMode, positionalArgs, JSON-Schema parameters).
- **controller**: `ensureSidecarToolsConfigMap` (immutable, GC'd, task-path only), read-only mount + `SIDECAR_TOOLS_MANIFEST_PATH` env, reserved volume name, oversize-manifest guard.
- **agent-runner**: trusted manifest loader; new `argv`/`stdin` exec-request mode (shell-free); `json.Number`-safe positional formatting; `--` end-of-options marker; runtime backstop against built-in/memory/MCP name collisions.
- **tool-executor.sh** (all skill sidecars): backward-compatible argv branch (base64-delimited so newline values stay one argument); legacy shell path untouched for `execute_command`.
- **webhook**: validation at both **AgentRun** and **SkillPack** admission (snake_case names, no built-in/memory collisions, cross-pack uniqueness, positionalArgs declared+required, args-mode params all positional).
- **pkg/sidecartools**: shared target normalization (de-dups controller/runner).
- **docs**: `writing-sidecars` / `writing-tools` updated with the security model + authoring caveats (flag injection, minimum executor version).

## Testing

- Unit tests at every layer, incl. a cross-language shell guard asserting argv newline-safety.
- `go build`/`go vet`/`go test ./...` green; `gofmt` clean; helm CRDs in sync.
- **Deployed to a kind cluster** and run end-to-end against a local llama.cpp + Qwen3.6-35B model:
  - native tool `svc_greet` invoked by the model → argv dispatch → correct result;
  - `execute_command` via a skill sidecar still works (legacy shell path);
  - sequential + parallel subagent ensembles pass (ordering + namespace handling);
  - controller writes/immutably-mounts/GCs the manifest as designed.

## Notes for review

- This is the full **CRD path** discussed on #226 — happy to close #226 in favor of this, or land both behind agreement on the invariant.
- `images/skill-echotool` is a minimal **reference sidecar** used for e2e; it's source-only and intentionally **not** wired into the Makefile `IMAGES`/release build — let me know if you'd prefer it added or moved under `examples/`.
- The admission webhook is the standard one (off in the default local-kind install); the controller-side behavior is fully exercised regardless.


AlexsJones:
Companion to #226 (KSiig's original native-sidecar-tools PR). This is the CRD-path implementation discussed there; #226 stays open so we can compare and decide whether to land this and retire it, or keep both.

PR #242, opened 2026-06-25
chore: replace Bitnami image references with Soldevelo equivalents

jkondrat-sd:
## Why this change

Bitnami changed its container image distribution model on August 28, 2025, and public images no longer receive ongoing updates or security fixes.
[SolDevelo](https://hub.docker.com/u/soldevelo) images are free drop-in replacements built from [SolDevelo/containers](https://github.com/SolDevelo/containers), a fork of bitnami/containers that continues to provide actively maintained and publicly available images.

## Image replacements

- `bitnami/kubectl:latest` → [`soldevelo/kubectl:1.3.6`](https://hub.docker.com/r/soldevelo/kubectl)

### DIFF
--- a/docs/guides/writing-skills.md
+++ b/docs/guides/writing-skills.md
@@ -106,7 +106,7 @@ Create a Dockerfile at `images/skill-<name>/Dockerfile`:
 # images/skill-my-tool/Dockerfile
 
 # Multi-stage: grab the binary you need
-FROM bitnami/kubectl:1.31 AS kubectl
+FROM soldevelo/kubectl:1.36 AS kubectl
 
 # Minimal base image
 FROM alpine:3.20
--- a/charts/sympozium/templates/pre-delete-finalizers.yaml
+++ b/charts/sympozium/templates/pre-delete-finalizers.yaml
@@ -90,7 +90,7 @@ spec:
       restartPolicy: Never
       containers:
         - name: strip-finalizers
-          image: bitnami/kubectl:latest
+          image: soldevelo/kubectl:1.36
           command:
             - /bin/sh
             - -c
--- a/docs/concepts/lifecycle-hooks.md
+++ b/docs/concepts/lifecycle-hooks.md
@@ -69,12 +69,12 @@ spec:
         verbs: ["get", "list", "create", "delete"]
     preRun:
       - name: create-context
-        image: bitnami/kubectl:latest
+        image: soldevelo/kubectl:1.36
         command: ["kubectl", "create", "configmap", "run-context",
                   "--from-literal=started=$(date)"]
     postRun:
       - name: cleanup-context
-        image: bitnami/kubectl:latest
+        image: soldevelo/kubectl:1.36
         command: ["kubectl", "delete", "configmap", "run-context"]
 ```
 
@@ -136,12 +136,12 @@ spec:
         verbs: ["create", "delete", "get"]
     preRun:
       - name: create-config
-        image: bitnami/kubectl:latest
+        image: soldevelo/kubectl:1.36
         command: ["sh", "-c",
           "kubectl create configmap agent-scratch --from-literal=run=$AGENT_RUN_ID"]
     postRun:
       - name: delete-config
-        image: bitnami/kubectl:latest
+        image: soldevelo/kubectl:1.36
         command: ["kubectl", "delete", "configmap", "agent-scratch"]
 ```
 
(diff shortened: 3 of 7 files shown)


AlexsJones:
Thanks for this — the Bitnami move is well-timed given their image-distribution changes, and this covers both remaining `bitnami/kubectl` references (the pre-delete finalizer job and the lifecycle-hooks integration test). Diff is clean.

Two small things before merge:

1. **Image provenance.** `soldevelo/kubectl` is a less-established publisher than the official upstream. Worth a sentence in the PR on why this mirror was chosen — or, if the goal is just a non-Bitnami source, consider the official `registry.k8s.io/kubectl` image, which avoids depending on any third-party Docker Hub account.
2. **Pin the tag.** Both refs stay on `:latest`. For a finalizer job that runs during cluster teardown, an unpinned tag is a reproducibility/supply-chain risk — pinning to a specific minor (e.g. `:1.31`) or a digest would be safer.

Neither blocks the intent; happy to approve once the image choice has a line of rationale and the tag is pinned.

jkondrat-sd:
Thanks for the review!

I've addressed the tag pinning and updated all remaining bitnami/kubectl references in the repository to use soldevelo/kubectl:1.36.

Regarding image provenance: the intent wasn't simply to replace Bitnami with another registry, but to preserve the Bitnami container behavior using an actively maintained continuation of the original project.

soldevelo/kubectl is built from the public SolDevelo/containers repository, which is a fork of bitnami/containers created after Bitnami stopped publishing updated public images. The images are intended to be drop-in compatible with the Bitnami images while continuing to receive upstream version and security updates. The entire build process and Dockerfiles remain public, which was the primary reason for choosing it over switching to a different image implementation.

While registry.k8s.io/kubectl would certainly also be a valid choice, that would be a migration to a different image family rather than a direct replacement for the previously used Bitnami image.

docs/guides/writing-skills.md
```diff
@@ -106,7 +106,7 @@ Create a Dockerfile at `images/skill-<name>/Dockerfile`:
 # images/skill-my-tool/Dockerfile
 
 # Multi-stage: grab the binary you need
-FROM bitnami/kubectl:1.31 AS kubectl
+FROM soldevelo/kubectl:1.36 AS kubectl
```
AlexsJones, line 109: You updated `FROM` here, but the `COPY --from=kubectl /opt/bitnami/kubectl/bin/kubectl /usr/local/bin/kubectl` a few lines down (~line 122) still references `/opt/bitnami/`. If the soldevelo fork keeps the binary at `/opt/bitnami/...` this works as-is — please confirm and add a brief note; otherwise the COPY needs updating too.

AlexsJones (changes requested):
Thanks for tackling the Bitnami deprecation — the motivation is sound and pinning to `:1.36` instead of `:latest` is a nice improvement. I confirmed `soldevelo/kubectl:1.36` exists on Docker Hub (multi-arch, current patch `1.36.2`), so no broken references. A couple of things to resolve before merge:

**1. Incomplete rename in `writing-skills.md` (inline comment).** The `FROM` was updated but the `COPY --from=kubectl /opt/bitnami/kubectl/bin/kubectl ...` on the next line still points at `/opt/bitnami/`. Since soldevelo is a fork of bitnami/containers the path may genuinely still be `/opt/bitnami/...`, but please verify and, if so, leave a note — as written it reads like a half-done rename.

**2. PR description doesn't match the diff.** The body lists `→ soldevelo/kubectl:latest`, but the diff pins `:1.36` everywhere (which is the better choice). Please update the description so the changelog/reviewers aren't misled.

**Non-blocking:**
- `writing-skills.md` also bumps kubectl `1.31 → 1.36`, bundling a version bump into the registry swap — worth calling out in case the `1.31` pin was intentional.
- This shifts a supply-chain trust boundary (Broadcom/Bitnami → SolDevelo) for an image that runs with cluster RBAC (configmap create/delete, gate patching). Consider digest-pinning (`@sha256:…`) at least in `charts/sympozium/templates/pre-delete-finalizers.yaml`, since that runs in users' clusters.
- Please make sure the `response-gate-*` e2e specs and the lifecycle integration test pass on CI — that's the real proof the soldevelo image behaves identically.

docs/guides/writing-skills.md
jkondrat-sd: Yes, soldevelo/kubectl is a fork of Bitnami's container definitions, so it intentionally preserves the same filesystem layout and paths (including /opt/bitnami).

jkondrat-sd:
> Refer

Thanks for the review. I have updated the PR description. Regarding /opt/bitnami, this is expected by design, SolDevelo images are fork of Bitnami's and intentionally keep the same filesystem layout.

PR #245, opened 2026-06-25
feat(slack): per-message sender attribution for multi-agent Ensembles

### LINKED ISSUE #235: Slack channel: per-message sender attribution (username/icon) for multi-agent Ensembles
Slack supports per-message username+icon_url/icon_emoji on chat.postMessage (scope chat:write.customize). 
In a multi-agent Ensemble, every `*-channel-slack` pod shares one Slack app (one bot identity), and 
`channels/slack/main.go` sendMessage posts chat.postMessage with only channel+text(+thread_ts) . we don't have  sender identity so a shared thread can't show which agent is replying.

Each channel pod already knows its agent (INSTANCE_NAME) and the Ensemble carries a per-agent displayName, so one bot can post as distinct identities without standing up ten separate Slack apps.

**Proposal**

1. Add optional Username + IconURL/IconEmoji to channel.OutboundMessage `internal/channel/types.go`

2. Set username/icon_url (or icon_emoji) on the Slack payload in SlackChannel.sendMessage, sourced from the agent displayName.

3. Document the new required bot scope chat:write.customize.

mvanhorn:
## Summary

Adds optional per-message sender attribution (`username` + `icon_url`/`icon_emoji`) to the Slack channel so a single shared Slack bot can post as distinct agent identities in a multi-agent Ensemble.

## Why this matters

Per #235: in a multi-agent Ensemble every `*-channel-slack` pod shares one Slack app (one bot identity), so a shared thread can't show which agent is replying. `sendMessage` posts `chat.postMessage` with only channel + text (+ `thread_ts`). Slack supports per-message `username` + `icon_url`/`icon_emoji` under the `chat:write.customize` scope, so one bot can present distinct identities without standing up a separate Slack app per agent. This follows the 3-point proposal in the issue and keeps the new fields optional on `OutboundMessage` as requested in review.

## Changes

- `internal/channel/types.go` — add optional `Username`, `IconURL`, `IconEmoji` (all `json:",omitempty"`) to `OutboundMessage`. Non-Slack channels ignore them, so they are untouched.
- `channels/slack/main.go` — in `sendMessage`, set `username`, and `icon_url` or `icon_emoji` on the `chat.postMessage` payload only when the corresponding field is non-empty. `icon_url` takes precedence over `icon_emoji` when both are set (matching Slack). Thread routing is unchanged.
- `channels/slack/main_test.go` — new tests covering the attribution payload, `icon_url` precedence, the legacy no-attribution payload (back-compat), and thread routing with attribution present.
- `docs/concepts/channels.md` — document the new required `chat:write.customize` bot scope in the Slack Setup section.

This is the transport-layer plumbing the issue describes. Populating the fields from the per-agent displayName on the producer side (the channel router) is a natural follow-up; happy to do that in a stacked PR if you'd prefer it in the same change.

## Testing

- `gofmt -l` clean on touched files
- `go vet ./internal/channel/... ./channels/slack/...` clean
- `go build ./...` succeeds
- `go test ./...` passes (Slack package: new tests green alongside existing ones)

Fixes #235

### DIFF
--- a/channels/slack/main.go
+++ b/channels/slack/main.go
@@ -596,6 +596,16 @@ func (sc *SlackChannel) sendMessage(ctx context.Context, msg channel.OutboundMes
 	if threadTS != "" {
 		payload["thread_ts"] = threadTS
 	}
+	// Per-message sender attribution: a single Slack bot can post as distinct
+	// per-agent identities via username + icon (requires chat:write.customize).
+	if msg.Username != "" {
+		payload["username"] = msg.Username
+	}
+	if msg.IconURL != "" {
+		payload["icon_url"] = msg.IconURL
+	} else if msg.IconEmoji != "" {
+		payload["icon_emoji"] = msg.IconEmoji
+	}
 	return sc.callSlackAPI(ctx, "https://slack.com/api/chat.postMessage", payload)
 }
 
--- a/channels/slack/main.go
+++ b/channels/slack/main.go
@@ -15,6 +15,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"flag"
 	"fmt"
 	"io"
@@ -536,6 +537,20 @@ type slackAPIResponse struct {
 	Error string `json:"error,omitempty"`
 }
 
+type slackAPIError struct {
+	Endpoint string
+	Code     string
+}
+
+func (e *slackAPIError) Error() string {
+	return fmt.Sprintf("slack %s rejected request: %s", e.Endpoint, e.Code)
+}
+
+func isSlackError(err error, code string) bool {
+	var slackErr *slackAPIError
+	return errors.As(err, &slackErr) && slackErr.Code == code
+}
+
 // callSlackAPI performs a JSON POST to the given Slack Web API endpoint
 // and returns an error when either the transport fails, the HTTP status
 // is non-2xx, or Slack reports ok:false. Errors classified as benign
@@ -575,11 +590,12 @@ func (sc *SlackChannel) callSlackAPI(ctx context.Context, endpoint string, paylo
 			return nil
 		}
 	}
-	return fmt.Errorf("slack %s rejected request: %s", endpoint, parsed.Error)
+	return &slackAPIError{Endpoint: endpoint, Code: parsed.Error}
 }
 
 // sendMessage sends a message via the Slack chat.postMessage API.
 func (sc *SlackChannel) sendMessage(ctx context.Context, msg channel.OutboundMessage) error {
+	const endpoint = "https://slack.com/api/chat.postMessage"
 	payload := map[string]interface{}{
 		"channel": msg.ChatID,
 		"text":    msg.Text,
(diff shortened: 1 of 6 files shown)


AlexsJones (approved):
Approving — verified this locally rather than just reading the diff: checked out the branch, `go build` and `go test ./channels/slack/` pass, the new tests correctly reuse the `newTestSlackChannel`/`jsonResponse` helpers from `reactions_test.go`, and the `icon_url` > `icon_emoji` precedence matches Slack's documented behavior. The fields being additive `omitempty` keeps old channel pods compatible with newer controllers, and `chat.postMessage` is the only text-posting path, so coverage is complete.

One thing to carry into the producer follow-up (non-blocking here since nothing populates the fields yet): once the router starts setting `Username`/icon, an app that hasn't been reinstalled with `chat:write.customize` will get `missing_scope` back from `chat.postMessage`, and the outbound loop logs and drops the message — the agent's reply is silently lost. Please make graceful degradation part of that stacked PR: on `missing_scope`, retry once without the attribution fields (falling back to the bot's default identity) and log a warning.

Also note this PR's `Fixes #235` will close the issue on merge even though the producer side isn't wired yet — fine by me as long as the follow-up lands; otherwise consider softening it to `Refs #235`.

AlexsJones (commented):

Automated multi-angle review; findings adversarially verified against this branch. Small, well-tested PR — but two of the findings are worth resolving before merge.

1. **Agent impersonation via the IPC path.** `internal/ipc/bridge.go` `handleOutboundMessage` relays agent-written JSON verbatim (`json.RawMessage(data)` → publish; no unmarshal, no field filtering), and the Slack channel unmarshals that raw payload into `channel.OutboundMessage`. Once this merges, any agent (e.g. prompt-injected) can write `{"channel":"slack","text":"…","username":"Alex Jones (CEO)","iconUrl":"https://…"}` to `/ipc/messages/send-*.json` and the shared bot posts under that identity — arbitrary display-name/icon spoofing of humans or other agents (Slack's APP badge remains, but that's thin protection). Suggest validating/allowlisting attribution at the bridge or a router chokepoint (e.g. only identities derived from the agent's own `AgentConfig.DisplayName`).

2. **`channels/slack/main.go:601` — no fallback when the workspace lacks `chat:write.customize`.** With `username`/icon set, `chat.postMessage` fails with `missing_scope`, `handleOutbound` only logs, and the reply is dropped entirely. Latent today (nothing populates the fields yet), but the moment a producer sets `Username`, every deployment that hasn't reinstalled the Slack app loses all attributed replies. Suggest: retry once without the attribution keys on `missing_scope` (best-effort presentation), or probe granted scopes at startup and disable attribution.

3. **⚠️ Cross-PR conflict with #258.** Both PRs append fields at the same spot in `internal/channel/types.go` (duplicate `IconEmoji` — a compile error on naive union — plus `Username` here vs `DisplayName` there, both feeding Slack's `username` key) and both edit the same region of `sendMessage` with different precedence/fallback behavior. The two need to be reconciled before either merges.

4. **`internal/ipc/protocol.go:177` — struct drift.** `ipc.OutboundMessage`'s doc comment says field names align with `channel.OutboundMessage` so the bridge can relay directly, but the three new fields weren't added there. The bridge's raw relay hides this today, but the documented contract is now further out of sync.

5. **`channels/slack/main.go:604` — icon precedence is inverted relative to Slack's documented semantics** (Slack: `icon_emoji` overrides `icon_url`; this code sends `icon_url` and drops `icon_emoji`). If deliberate, worth a comment saying so — `TestSendMessage_IconURLTakesPrecedence` locks it in.

6. **`internal/channel/types.go:44` — Slack-specific fields on the channel-agnostic type.** `Metadata` is the documented channel-specific extension point (Slack `replyToTS` uses it today). Alternatively, since #235's problem (which agent is replying) is cross-channel, a channel-neutral sender-identity (name + avatar) that each channel maps natively would fix it for Discord/Telegram/WhatsApp too instead of Slack only.

Nit: the four new tests each repeat the same ~15-line capture/send/decode scaffold; a `postAndDecode(t, msg)` helper or a table-driven test would halve the file.

mvanhorn:
Both findings addressed in 58a5873:

1. IPC attribution allowlist: `handleOutboundMessage` now sanitizes agent-written payloads before publish. `username` must match the agent's own configured display name or it is stripped, `iconUrl`/`iconEmoji` are dropped at the bridge, and dropped fields are logged. Spoofed identities never reach the Slack channel.

2. missing_scope fallback: `chat.postMessage` now retries once without the attribution fields when Slack returns `missing_scope`, so replies fall back to the bot's default identity instead of being dropped, with a warning logged that attribution was degraded.

Added tests for both paths. `go build`, `go vet`, and `go test ./internal/ipc/... ./channels/slack/...` pass locally.

mvanhorn:
Thank you @AlexsJones for landing per-message sender attribution.

mvanhorn:
Thanks @AlexsJones! Happy that per-message sender attribution is flowing through the Slack integration now.

PR #258, opened 2026-07-02
feat(ensemble): designate a Slack receiver persona and delegate on @mention

### LINKED ISSUE #257: [Feature] Ensemble: designate a Slack receiver persona and delegate on @mention
### Problem
​
When an Ensemble is bound to Slack, an inbound message is currently routed to a
single agent using a hardcoded `AgentID: "primary"`
(`internal/controller/channel_router.go`). There is no declarative way to say
*which* persona in the crew should receive Slack messages, and there is no way
for a user to address a specific teammate from within a message. The result:
you can't tell who in the ensemble actually receives a message, and you can't
target the right agent for a given request.
​
### Proposed feature
​
Two additions, both building on primitives that already exist in the CRD:
​
**1. Designate the Slack receiver.**
Add an optional `slackListener: bool` field to `AgentConfigSpec`
(`api/v1alpha1/ensemble_types.go`). The persona with `slackListener: true` is
the default receiver for inbound Slack messages on the ensemble's Slack channel.
​
- Exactly one persona per ensemble should set it.
- If none set it, behaviour is unchanged (first Slack-bound persona), so this is
  backwards compatible.
​
**2. Delegate on `@mention`.**
Agents already carry `Name` and `DisplayName` (`AgentConfigSpec`). When an
inbound message contains `@<name>` (matching `Name` or `DisplayName`,
case-insensitive), the designated receiver **delegates** the task to that named
persona rather than handling it itself. The receiver stays the front door;
delegation flows through the existing delegation executor
(`Relationships` / `WorkflowType: delegation`), so the guardrails (in-flight and
depth caps) and delegation telemetry apply automatically. An unrecognised name
leaves the message with the receiver plus a friendly note.
​
### Example
​
```yaml
kind: Ensemble
spec:
  workflowType: delegation
  agentConfigs:
    - name: lead
      displayName: Lead
      slackListener: true          # <- receives inbound Slack messages
      channels: ["slack"]
    - name: researcher
      displayName: Researcher
      channels: ["slack"]
  relationships:
    - from: lead
      to: researcher
      type: delegation
```
​
A Slack message `"@Researcher can you dig into X?"` lands on `lead`, which
delegates the task to `researcher`.
​
### Existing building blocks (keeps this incremental)
​
- `AgentConfigSpec.Name` / `.DisplayName` already exist.
- The delegation executor (`Relationships`, `WorkflowType: delegation`) already
  runs, with in-flight/depth-cap guardrails.
- Prior unmerged work parses `@persona` / `persona:` in inbound channel messages
  and resolves the sibling Agent CR — reusable for the `@mention` parse/resolve
  step.
​
### Acceptance criteria
​
- [ ] `slackListener` field added to `AgentConfigSpec`; CRDs regenerated.
- [ ] Inbound Slack messages route to the `slackListener` persona; fallback
      preserved when unset.
- [ ] `@name` (Name or DisplayName) in an inbound message delegates from the
      receiver to the named persona via the delegation executor.
- [ ] Unknown `@name` → stays on receiver, no crash, user-visible note.
- [ ] Outbound replies from a delegated agent are attributed to that agent.
- [ ] Unit tests + docs (`docs/concepts/channels.md`) + a sample Ensemble YAML.
​
### Out of scope
​
- Multiple Slack workspaces/bots per ensemble (today: one ensemble → one Slack
  deployment).
- Per-persona Slack visual identity (icon/name), tracked separately.
- Non-Slack channels (Discord already supports per-channel routing via
  `AllowedChats`).

henrikrexed:
## Summary

Allows an Ensemble to designate one persona as the Slack receiver via a `slackListener` boolean flag on `AgentConfigSpec`, and re-routes inbound messages addressed with `@<name>` to the named persona rather than the receiver.

Closes #257

### CRD
- `AgentConfigSpec.SlackListener bool` (`json:"slackListener,omitempty"`) — marks exactly one persona as the inbound Slack receiver
- CEL `XValidation` enforces at most one `slackListener: true` per Ensemble; a warning log is emitted when the constraint is violated so the operator surfaces misconfigurations without hard-failing
- `sympozium.ai/slack-listener` label + annotation stamped on the receiver's Agent CR; self-healed by `setSlackListenerMetadata` on every reconcile

### Channel router
- `resolveSlackReceiver` — scans `AgentConfigs` for the first `SlackListener: true` entry and swaps the AgentRun target to that persona's Agent CR; falls back to the inbound-receiving pod when no listener is configured
- `extractNameMention` — scans the message for the first boundary-anchored `@<token>` anywhere in the text (not leading-only); email addresses, URLs, and Slack keywords (`@here`, `@channel`, `@everyone`) are excluded
- `resolveNamedDelegate` — matches the extracted mention against each persona's `Name` and `DisplayName` (case-insensitive); first match wins
- `handleInbound`: `@name` routing checked first; if matched, AgentRun is created for the named delegate; if the mention exists but matches no persona and the channel is not muted/inaccessible, a denial is sent back to Slack; unaddressed messages fall through to `resolveSlackReceiver`

### Outbound attribution
- `agentDisplayName` helper reads the `sympozium.ai/agent-display-name` annotation from the sending Agent CR
- `chat.postMessage` now sets `username` to the resolved display name so Slack shows which persona replied

### Tests
- `TestExtractNameMention` — 13 table-driven cases covering mid-message mention, email/URL exclusion, Slack keywords, case variants
- `TestResolveNamedDelegate` — 7 cases: exact Name, case-insensitive Name, DisplayName, partial-no-match, unknown-nil, empty-nil
- `TestHandleInbound_NameRouting` — 4 integration cases: known mention creates delegate AgentRun, unknown mention sends denial, non-mention prefix falls through to receiver, unaddressed falls through to listener swap
- `TestBuildAgent_SlackListenerStamp` — listener vs non-listener label/annotation stamping
- `TestSetSlackListenerMetadata` — 6 self-heal subtests

## Compatibility

- No changes to existing AgentRun spec, Ensemble spec fields other than the additive `slackListener` flag
- Ensembles without `slackListener: true` on any persona continue to route to the inbound-receiving pod (backward compatible)
- `@name` routing requires the inbound message to arrive at a pod that belongs to an Ensemble; standalone Agents are unaffected

### DIFF
--- a/api/v1alpha1/ensemble_types.go
+++ b/api/v1alpha1/ensemble_types.go
@@ -288,6 +288,16 @@ type AgentConfigSpec struct {
 	// When empty, the provider-appropriate default applies.
 	// +optional
 	RunTimeout string `json:"runTimeout,omitempty"`
+
+	// SlackListener marks this persona as the designated receiver for inbound
+	// Slack messages. At most one persona per Ensemble should set this to true;
+	// if none do, the channel router falls back to the first Slack-bound persona
+	// (today's behaviour). The ensemble-controller propagates this flag as the
+	// annotation "sympozium.io/slack-listener: true" on the generated Agent CR
+	// so the channel router can resolve the receiver without reading the Ensemble.
+	// +optional
+	// +kubebuilder:validation:Optional
+	SlackListener bool `json:"slackListener,omitempty"`
 }
 
 // AgentConfigWebEndpoint configures the web endpoint for an agent configuration.
--- a/api/v1alpha1/ensemble_types.go
+++ b/api/v1alpha1/ensemble_types.go
@@ -28,6 +28,11 @@ type EnsembleSpec struct {
 	Version string `json:"version,omitempty"`
 
 	// Personas is the list of agent personas in this ensemble.
+	// At most one persona may set slackListener=true — the designated Slack
+	// receiver (ISI-1497). Setting it on more than one persona is rejected at
+	// admission so the misconfiguration surfaces instead of silently resolving
+	// to the first declared listener.
+	// +kubebuilder:validation:XValidation:rule="self.filter(c, has(c.slackListener) && c.slackListener).size() <= 1",message="at most one persona may set slackListener=true (the designated Slack receiver)"
 	AgentConfigs []AgentConfigSpec `json:"agentConfigs"`
 
 	// AuthRefs references secrets containing AI provider credentials.
@@ -290,11 +295,14 @@ type AgentConfigSpec struct {
 	RunTimeout string `json:"runTimeout,omitempty"`
 
 	// SlackListener marks this persona as the designated receiver for inbound
-	// Slack messages. At most one persona per Ensemble should set this to true;
-	// if none do, the channel router falls back to the first Slack-bound persona
-	// (today's behaviour). The ensemble-controller propagates this flag as the
-	// annotation "sympozium.io/slack-listener: true" on the generated Agent CR
-	// so the channel router can resolve the receiver without reading the Ensemble.
+	// Slack messages. At most one persona per Ensemble may set this to true
+	// (enforced by CEL validation on agentConfigs); if none do, the channel
+	// router falls back to the first Slack-bound persona (today's behaviour).
+	// The channel router resolves the receiver by reading the Ensemble CR
+	// (see resolveSlackReceiver). The ensemble-controller additionally stamps
+	// the label/annotation "sympozium.ai/slack-listener: true" on the generated
+	// Agent CR for observability and kubectl selectability
+	// (`kubectl get agents -l sympozium.ai/slack-listener=true`).
 	// +optional
 	// +kubebuilder:validation:Optional
 	SlackListener bool `json:"slackListener,omitempty"`
--- a/channels/slack/main.go
+++ b/channels/slack/main.go
@@ -15,6 +15,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"flag"
 	"fmt"
 	"io"
@@ -536,6 +537,18 @@ type slackAPIResponse struct {
 	Error string `json:"error,omitempty"`
 }
 
+// slackAPIError carries the machine-readable Slack error code from an
+// ok:false response so callers can branch on the exact code rather than
+// substring-matching the formatted message (ISI-1561 #4).
+type slackAPIError struct {
+	Endpoint string
+	Code     string
+}
+
+func (e *slackAPIError) Error() string {
+	return fmt.Sprintf("slack %s rejected request: %s", e.Endpoint, e.Code)
+}
+
 // callSlackAPI performs a JSON POST to the given Slack Web API endpoint
 // and returns an error when either the transport fails, the HTTP status
 // is non-2xx, or Slack reports ok:false. Errors classified as benign
@@ -575,7 +588,7 @@ func (sc *SlackChannel) callSlackAPI(ctx context.Context, endpoint string, paylo
 			return nil
 		}
 	}
-	return fmt.Errorf("slack %s rejected request: %s", endpoint, parsed.Error)
+	return &slackAPIError{Endpoint: endpoint, Code: parsed.Error}
 }
 
 // sendMessage sends a message via the Slack chat.postMessage API.
@@ -603,15 +616,40 @@ func (sc *SlackChannel) sendMessage(ctx context.Context, msg channel.OutboundMes
 		payload["icon_emoji"] = msg.IconEmoji
 	}
 	err := sc.callSlackAPI(ctx, "https://slack.com/api/chat.postMessage", payload)
-	if err != nil && strings.Contains(err.Error(), "not_allowed_token_type") {
-		// chat:write.customize scope missing — retry without customization fields.
+	// Retry without the chat:write.customize fields when the bot token cannot
+	// customize the sender. Classic tokens report not_allowed_token_type;
+	// granular bot tokens missing the scope report missing_scope. Match the
+	// machine-readable code from the typed error rather than substring-scanning
+	// the formatted message (ISI-1561 #4).
+	if isMissingCustomizeScope(err) {
 		sc.log.Info("bot token lacks chat:write.customize; sending without username/icon override",
-			"displayName", msg.DisplayName)
+			"displayName", msg.DisplayName, "code", slackErrorCode(err))
 		delete(payload, "username")
 		delete(payload, "icon_emoji")
 		err = sc.callSlackAPI(ctx, "https://slack.com/api/chat.postMessage", payload)
 	}
 	return err
+}
+
+// slackErrorCode returns the machine-readable Slack error code carried by a
+// *slackAPIError, or "" when err is nil or not a Slack API error.
+func slackErrorCode(err error) string {
+	var apiErr *slackAPIError
+	if errors.As(err, &apiErr) {
+		return apiErr.Code
+	}
+	return ""
+}
+
+// isMissingCustomizeScope reports whether err indicates the bot token cannot
+// set username/icon overrides (the chat:write.customize scope is absent).
+func isMissingCustomizeScope(err error) bool {
+	switch slackErrorCode(err) {
+	case "not_allowed_token_type", "missing_scope":
+		return true
+	default:
+		return false
+	}
 }
 
 // addReaction adds an emoji reaction to a message via the Slack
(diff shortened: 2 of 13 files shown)


AlexsJones (commented):

Automated multi-angle review; every finding below was adversarially verified against this branch with quoted code.

### Correctness

1. **`internal/controller/channel_router.go:303` — @name swap bypasses access control and mute.** The delegate swap (`inst = &delegateInst`) runs *before* `checkChannelAccess` and `applyTriggers`. `checkChannelAccess` returns allow-all when the instance has no matching channel spec (typical for delegate personas), and mute state is keyed per-instance (`<inst>-channel-state`), so a sender blocked by the receiver's `allowedSenders` — or a chat muted via a stop keyword — gets an AgentRun created by sending `"@billing do X"`. The listener-fallback swap (~line 334) has the same issue in reverse: the arriving instance's access rules are dropped in favor of the listener's (possibly absent) rules.

2. **`internal/controller/channel_router.go:295,335` — routing not gated on `msg.Channel == "slack"`.** `handleInbound` is the single subscriber for all channels (telegram/whatsapp/discord too), and both new blocks check only the ensemble label. A Telegram message `"ask @john about the deploy"` gets an unknown-persona denial and is dropped; a WhatsApp message gets rerouted to the *Slack* listener persona.

3. **`internal/controller/channel_router.go:681` — punctuation and casual @words drop messages.** `extractNameMention` splits only on whitespace (`strings.IndexAny(rest, " \t\n\r")`), so `"@billing, can I get a refund?"` extracts token `billing,`, fails exact matching, and the message is denied and dropped. Same for any non-persona `@word` anywhere in the text (`"I'll be @home tomorrow"`).

4. **`channels/slack/main.go:606` — replies silently dropped on workspaces without `chat:write.customize`.** The fallback retry matches only `not_allowed_token_type`, but granular bot tokens return `missing_scope`. Since `agentDisplayName` never returns empty, **every** completed-run reply now sets `username`, and `handleOutbound` only logs send errors — so existing deployments that haven't reinstalled the app with the new scope lose all agent replies. Match `missing_scope` too (ideally surface the parsed Slack error code from `callSlackAPI` instead of substring-matching), and cache the discovered lack of scope so it's not two API calls per message forever.

5. **`internal/controller/channel_router.go:379` — unknown-mention denial fails open.** `if ensErr == nil { …deny…; return }` has no else/return, so a transient Get error falls through and creates an AgentRun with the unstripped @mention text — nondeterministic behavior for the same input.

6. **`internal/controller/channel_router.go:305` — bare `"@persona"` creates an AgentRun with an empty Task.** The empty-text guard runs before the remainder is assigned to `msg.Text`; nothing re-checks afterwards.

### Cleanup

7. **Up to four `client.Get`s of the same Ensemble per inbound message** (@name block ~287, listener block ~327, denial block ~379, plus one inside `agentDisplayName` ~427), all with the same key. Fetch once at the top and thread the `*Ensemble` through — this also removes torn reads between the routing decision, denial text, and stamped display name.

8. **`internal/controller/channel_router.go:300,331` — instance names rebuilt by string concatenation** (`ensembleName + "-" + delegate.Name`), duplicating the ensemble controller's naming convention. `Ensemble.Status.InstalledAgentConfigs` (already used by `spawner.resolvePersonaTarget`) is the authoritative persona→instance mapping, and the router already holds the fetched Ensemble at both sites.

Minor: `DisplayName` is serialized twice per completed-run event (`OutboundMessage.DisplayName` + `eventMeta["agentDisplayName"]`, which nothing reads), and `IconEmoji` is plumbed and tested but no producer ever sets it.

⚠️ **Cross-PR note:** #245 adds `Username`/`IconURL`/`IconEmoji` to the same `OutboundMessage` struct region and the same `sendMessage` area — duplicate `IconEmoji` field and two different field names (`DisplayName` vs `Username`) feeding the same Slack `username` key. The two PRs need to be reconciled before either merges.

henrikrexed:
Thanks for the review, @AlexsJones — all eight findings are addressed. Pushed as `db9de1d`.
​
The core change restructures `handleInbound` in `channel_router.go`: the Ensemble is now fetched **once** at the top, and the flow cleanly separates the **front door** (the Slack-facing persona that owns the channel and its access rules) from the **compute target** (the `@name` delegate). Access is always evaluated on the front door; the delegate only ever receives already-authorized work.
​
### Correctness
​
**1 — `@name` swap bypassed `allowedSenders` + per-instance mute (`:303`)**
Fixed. Access and mute are now evaluated against the front door instance, never the delegate. A sender who is blocked on the listener can no longer reach a delegate via `@billing …`. Covered by new test `TestHandleInbound_FrontDoorAccessControl`.
​
**2 — routing not gated on channel type (`:295,335`)**
Fixed. Both the `@name` block and the SlackListener block are now guarded by `msg.Channel == "slack"`, so Telegram/WhatsApp/Discord messages skip Slack-specific rerouting entirely. New test `TestHandleInbound_NonSlackSkipsRouting` (a Telegram `@john` is neither denied nor rerouted).
​
**3 — `extractNameMention` split only on whitespace (`:681`)**
Fixed. The extracted token is run through `strings.TrimRight(token, ",.;:!?")` before matching, so `@billing,` now matches. Four new cases added to `TestExtractNameMention`.
​
**4 — reply fallback matched only `not_allowed_token_type` (`channels/slack/main.go:606`)**
Fixed. Introduced a typed `*slackAPIError{Code}` and switched the fallback to `errors.As` on the machine-readable Slack error code (no substring scan). `sendMessage` now retries on both `not_allowed_token_type` and `missing_scope`, so granular bot tokens are handled.
​
**5 — unknown-mention denial failed OPEN on a transient Get error (`:379`)**
Fixed. Eliminated the redundant second `Ensemble.Get`; the denial now uses the single top-of-function fetch, so a transient error can no longer fall through into an AgentRun carrying unstripped `@mention` text.
​
**6 — bare `@persona` created an AgentRun with empty Task (`:305`)**
Fixed. `msg.Text` is re-checked for emptiness *after* the mention prefix is stripped and the remainder reassigned. A message that is nothing but `@persona` no longer spawns an empty-Task run. New test `TestHandleInbound_EmptyTaskAfterMention`.
​
### Cleanup
​
**7 — up to 4 `client.Get` of the same Ensemble per inbound message**
Fixed. The Ensemble is fetched once at the top and threaded through every routing decision, removing the torn-read window.
​
**8 — instance names rebuilt by string concat (`:300,331`)**
Fixed. New helper `personaInstanceName()` resolves the instance via `Ensemble.Status.InstalledAgentConfigs` — the same authoritative map used by `spawner.resolvePersonaTarget` — falling back to `<ensemble>-<persona>` only during the brief pre-install window.
​
### Minor / cross-PR
​
- **`DisplayName` double-serialization** and the **unset `IconEmoji` producer**: acknowledged; left as-is since neither affects behavior.
- **PR #245 overlap**: #245 adds `Username`/`IconURL`/`IconEmoji` to the same `OutboundMessage` region, which duplicates the `IconEmoji` field and gives two names (`DisplayName` vs `Username`) feeding the same Slack `username` key. This needs reconciliation with the #245 owner before either merges — flagging for coordination; out of scope for this remediation.
​
### Verification (local)
`gofmt` clean · `go vet ./internal/controller ./channels/slack` clean · `go test ./internal/controller ./channels/slack` PASS, including the 3 new integration tests and 4 new punctuation cases.

AlexsJones (changes requested):
Reviewed the full diff — this is a strong, carefully-built PR. The security posture stands out: access-control and mute are evaluated against the resolved *front door* (never an `@name` delegate), which correctly closes the `@billing do X` bypass; the Ensemble is fetched once and threaded through; non-Slack channels skip routing entirely; and the empty-task-after-strip guard, punctuation trimming, boundary-anchored `@` (email/URL-safe), and Slack-keyword exclusion are all handled, with 900+ lines of table-driven tests. Nice work @henrikrexed.

One change before this is ready to merge — a doc/validation contradiction that will mislead users:

- The CRD ships a **hard** CEL admission rule: `self.filter(c, has(c.slackListener) && c.slackListener).size() <= 1`. That *rejects* an Ensemble with two listeners at `kubectl apply` time.
- But `docs/concepts/channels.md` says setting it on more than one persona "triggers a kubebuilder validation warning; only the first match is used at runtime", and the `SlackListener` godoc / PR description say the constraint is surfaced "without hard-failing" via a warning log. `resolveSlackReceiver`'s first-wins logic also implies multiples are tolerated at runtime.

These describe two mutually exclusive behaviours. The `EnsembleSpec.AgentConfigs` godoc gets it right ("rejected at admission"). Please reconcile to one story: either keep the hard CEL rule (recommended — it matches the stated intent) and fix `channels.md` + the field godoc + PR body to say the apply is rejected (there is no runtime "first wins"), or, if soft/first-wins is genuinely wanted, drop the CEL rule to a warning-only mechanism. Most likely a small docs fix.

Everything else looks ready for @AlexsJones's final look.

🤝 _Handled by Alex's repo steward — replies reviewed by [@AlexsJones](https://github.com/AlexsJones)_

PR #264, opened 2026-07-08
fix: align Ensemble update path with create path for all Agent fields

KSiig:
## Summary
- **AuthRefs**: update path now filters by `persona.Provider`, matching create path logic. Previously passed all refs unfiltered, causing agents to receive the wrong API key when multiple providers were configured.
- **Model**: update path now applies the same default/override logic as create — defaults to `gpt-4o` when `persona.Model` is empty, and overrides to `pack.Spec.ModelRef` when a local model endpoint is active.
- **Volumes/VolumeMounts**: now propagated on update (were set at creation but never synced).
- **AgentSandbox**: now propagated on update.
- **Lifecycle**: now propagated on update.
- **PolicyRef**: now propagated on update.

Extracts `resolveAuthRefs()` and `resolveModel()` helpers shared by both create and update paths to prevent future divergence.

## Root cause
The create path (`buildAgent`) and update path (field-by-field comparison in `reconcileAgentConfig`) were maintained independently. New fields added to `buildAgent` were not always backfilled into the update path, and existing fields (authRefs, model) used different logic.

## Test plan
- [x] Unit tests for `resolveAuthRefs` (provider filtering, local endpoint, no provider)
- [x] Unit tests for `resolveModel` (persona override, default, local endpoint)
- [x] Update propagation test: authRefs filtered to matching provider
- [x] Update propagation test: volumes and volume mounts
- [x] Update propagation test: sandbox, lifecycle, and policy ref
- [x] Update propagation test: model override for local endpoints + auth cleared
- [x] Full controller test suite passes with no regressions

### DIFF
--- a/internal/controller/ensemble_controller.go
+++ b/internal/controller/ensemble_controller.go
@@ -321,15 +321,17 @@ func (r *EnsembleReconciler) reconcileAgentConfig(
 			needsUpdate = true
 		}
 
-		// Propagate authRefs changes.
-		if !authRefsEqual(existingInst.Spec.AuthRefs, pack.Spec.AuthRefs) {
-			existingInst.Spec.AuthRefs = pack.Spec.AuthRefs
+		// Propagate authRefs changes (filtered by persona provider).
+		wantAuthRefs := resolveAuthRefs(pack, persona, modelEndpoint)
+		if !authRefsEqual(existingInst.Spec.AuthRefs, wantAuthRefs) {
+			existingInst.Spec.AuthRefs = wantAuthRefs
 			needsUpdate = true
 		}
 
-		// Propagate model changes from persona definition.
-		if persona.Model != "" && existingInst.Spec.Agents.Default.Model != persona.Model {
-			existingInst.Spec.Agents.Default.Model = persona.Model
+		// Propagate model changes (with same defaults as buildAgent).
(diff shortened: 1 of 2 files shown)


AlexsJones:
Pushed a follow-up commit (`1589546`) addressing two review points:

- **Mixed ensembles now work.** A persona that pins its own `provider` fully opts out of the ensemble's cluster-local model endpoint — it keeps its own `model` (not the ensemble `modelRef`), takes its matching cloud auth key (was previously nil'd under a local endpoint), and uses its own/base URL instead of the local endpoint. `persona.Provider != ""` is the single signal, applied consistently across `resolveModel` / `resolveAuthRefs` / `resolveBaseURL` via a shared `personaOverridesEndpoint()` helper.
- **Finished the create/update DRY.** Extracted `resolveBaseURL()` and `resolveProviderHeadersSecretRef()` so `buildAgent` (create) and `reconcileAgentConfig` (update) share the logic for every model field — no more inline duplication that could drift.

Added unit tests for `resolveBaseURL` and the mixed-ensemble cases, plus an end-to-end `reconcileAgentConfig` test asserting a cloud persona under a local-model ensemble gets its provider key, its own model, and a non-local base URL. `go test -race ./internal/controller/...`, `vet`, `gofmt` all clean.

PR #277, opened 2026-07-09
avoid AwaitingDelegate causing stacked schedule runs

kristof-ringleff:
The Forbid policy check did not include all run statuses that could be considered running. As a result scheduled runs would stack up if a run was in status AgentRunPhaseAwaitingDelegate.

### DIFF
--- a/internal/controller/sympoziumschedule_controller.go
+++ b/internal/controller/sympoziumschedule_controller.go
@@ -170,10 +170,7 @@ func (r *SympoziumScheduleReconciler) Reconcile(ctx context.Context, req ctrl.Re
 		); err == nil {
 			for i := range scheduleRuns.Items {
 				phase := scheduleRuns.Items[i].Status.Phase
-				if phase == sympoziumv1alpha1.AgentRunPhaseRunning ||
-					phase == sympoziumv1alpha1.AgentRunPhasePending ||
-					phase == sympoziumv1alpha1.AgentRunPhaseServing ||
-					phase == "" {
+				if isAgentRunActive(phase) {
 					log.Info("Skipping trigger — active run exists (Forbid policy)",
 						"activeRun", scheduleRuns.Items[i].Name, "phase", phase)
 					_ = r.Status().Update(ctx, schedule)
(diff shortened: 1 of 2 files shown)


AlexsJones (approved):
Reviewed and validated locally — **approve**. This is a correct, well-scoped fix.

## The bug is real and the fix is complete

The \`Forbid\` concurrency gate (and the parallel \`pipelineInFlight\` gate) enumerated only \`Pending\`/\`Running\`/\`Serving\`/\`""\` as "active", so an AgentRun parked in \`AwaitingDelegate\` (an orchestrator waiting on a \`delegate_to_persona\` child — potentially tens of minutes) or in \`PostRunning\` (postRun hooks still executing) read as finished, and the next cron tick stacked a second run on the live one.

I checked the full phase taxonomy in \`api/v1alpha1/agentrun_types.go\` — there are exactly 8 \`AgentRunPhase\` constants. The new \`isAgentRunActive\` partitions them completely and correctly:
- **active** → Pending, Running, Serving, PostRunning, AwaitingDelegate, \`""\`
- **terminal** → Succeeded, Failed, Skipped

The terminal three match the type comments precisely (\`Skipped\` is explicitly documented as terminal). Nothing is missed.

## Consolidation is the right call

Both gates had independently drifted — the \`pipelineInFlight\` switch was *also* missing PostRunning/AwaitingDelegate. Extracting one helper is exactly what prevents the two sites from diverging again, and the doc comment ("keep this in sync with the AgentRunPhase constants") flags the maintenance contract.

## Local validation

- \`go build\`, \`go vet\`, \`gofmt\` — clean
- \`go test -race ./internal/controller/\` — passes (the repo bar)
- **Regression-guard check**: I reverted just the production fix (kept the new test) and confirmed \`TestSympoziumScheduleReconcile_ForbidBlocksOnNonTerminalPhases\` fails on the \`AwaitingDelegate\` and \`PostRunning\` sub-cases, then passes once restored. The test genuinely exercises the fix rather than tautologically tracking it. Good table coverage across all non-terminal and terminal phases.

## Optional, non-blocking

- \`internal/controller/ensemble_controller.go:937\` has a similar active-phase switch (its disable-cancel path) that includes \`AwaitingDelegate\` but omits \`PostRunning\`/\`""\`. Different semantics (cancel-on-disable, not a concurrency gate), so not a correctness bug here — but it is now the last place that could adopt \`isAgentRunActive\` if you want a single source of truth for "run not finished". Fine to leave for a follow-up.

LGTM.

PR #278, opened 2026-07-09
Determine delegation timeout from remaining parent runtime

kristof-ringleff:
When an agent delegated to another run a hardcoded 10 minute timeout was in place for waiting for the result to come back via IPC irrespective of the runtime. The changes in this PR pass through the parent's deadline.

Also consolidated spawnSubagentsTool and delegateToPersonaTool to use the same ipcRoundTrip helper as they used the same mechanic.

While working on this I found the existing relationship timeout which previously existed in the schema but wasn't read/enforced anywhere. This PR uses that field if present in the first instance, otherwise the parent's deadline is applied.


AlexsJones (changes requested):
Thanks @kristof-ringleff — this is a genuinely nice piece of work. Deriving the delegate/subagent IPC wait from the parent run's remaining context deadline (instead of the hardcoded 10-minute wait), consolidating both paths onto one `ipcRoundTrip` helper over the `ipc.*` wire types, and finally giving `relationships[].timeout` real server-side enforcement is exactly the right direction. The deadline math and clamping are sound — `delegateWaitBudget` reads `ctx.Deadline()` off the propagated run timeout, a past deadline is rejected pre-flight with "no time left in the run budget", `ParseTimeout` cleanly rejects empty/zero/negative/malformed, and `expireDelegation` coordinates exactly-once via `LoadAndDelete`. Test coverage of all of that is good.

There's one blocking defect to fix before this can go in, though:

### 🔴 Blocking: `spawn_subagents` will panic the router goroutine

The refactor converts the **delegation** store site to a pointer (`spawn_router.go:229`, `sr.pending.Store(name, pd)` where `pd := &pendingDelegation{...}`) and updates the retrieval sites to assert the pointer type:

- `handleChildCompleted` — `pd := val.(*pendingDelegation)` (line 274)
- `handleChildFailed` — `pd := val.(*pendingDelegation)` (line 321)
- `expireDelegation` — `pd := val.(*pendingDelegation)` (line 1135)

But the two **subagent-batch** store sites into the same `sr.pending` map were not converted and still store the struct **by value**:

- parallel batch — `spawn_router.go:509` — `sr.pending.Store(result.RunName, pendingDelegation{...})`
- sequential batch — `spawn_router.go:761` — `sr.pending.Store(result.RunName, pendingDelegation{...})`

`handleChildCompleted`/`handleChildFailed` run for *every* child completion — they `LoadAndDelete` from `sr.pending` and assert `*pendingDelegation` at line 274/321 **before** dispatching to `handleBatchChildDone` at the end. So the first time any subagent batch child completes or fails, `val.(*pendingDelegation)` fails on a `pendingDelegation` value and panics on the SpawnRouter event-loop goroutine — `spawn_subagents` is effectively non-functional and the controller goroutine crashes.

**Fix:** convert lines 509 and 761 to `&pendingDelegation{...}` as well. Better still, funnel all three writers through a single `storePending(name, pd)` helper (or make `pending` a typed wrapper) so the value/pointer contract can't drift again.

### Test gap that hides it
Every new fixture stores `&pendingDelegation{}` (pointer), so none exercise a value-stored subagent child through `handleChildCompleted`/`handleChildFailed` — which is exactly why CI is green despite the panic. Please add a regression test that drives a subagent-batch child (as actually stored) through completion, so this seam is covered.

### Minor (non-blocking)
- When a declared edge timeout exceeds the remaining parent runtime, the client backstop unblocks the parent first, yet the server-side `expireDelegation` still fires later (child delete + circuit-breaker increment) for a parent that already gave up. Worth a guard or a comment so it isn't mistaken for a spurious failure.
- `ipcRoundTrip` intentionally leaves the request file on timeout/cancel (router may still act) but never reaps a late-written result file — pre-existing orphan-file pattern, just noting it.

Really solid overall — fix the pointer/value inconsistency and the accompanying test and I think this is close.

kristof-ringleff:
The issues should be addressed now including additional test coverage for the failure.

AlexsJones (approved):
Delta re-review of the three commits since my last pass — everything requested has been addressed, and thoroughly. ✅

- **Blocking pointer/value defect**: fixed exactly as suggested — all three writers now funnel through a `storePending` helper whose doc comment states the `*pendingDelegation` contract, and both subagent-batch sites (parallel and sequential) store pointers. The seam can't silently drift again.
- **Regression coverage**: the new `spawn_router_subagent_test.go` drives a parallel batch through `handleSubagentRequest` → `handleChildCompleted`/`handleChildFailed` using the production store path (one child succeeding, one failing), which is precisely the test that would have caught the panic.
- **Settled-parent guard** (the minor note — you went further than asked): `expireDelegation` now treats a Succeeded/Failed/Skipped or deleted parent as cleanup-only — orphaned child reaped, but no ghost delegate-failure publish, no status rewrite, no circuit-breaker increment — with `TestExpireDelegation_ParentAlreadySettled` and `TestExpireDelegation_ParentDeleted` locking it in. The companion gate in `updateParentDelegateStatus` (only resurrect to Running from `AwaitingDelegate`) closes the phase-resurrection hole from the other side, and the follow-up refactor to `isAgentRunActive` keeps the phase set in one place.

CI is green and the branch is current with main. Approving — recommending to @AlexsJones for final look and merge. Nice work on this one.

PR #289, opened 2026-07-13
feat: ToolPolicy allow/deny propagation and runner-side filtering

KSiig:
## Summary
- Wire `toolPolicy.deny` and `toolPolicy.allow` from Ensemble `agentConfigs[]` through to AgentRun pods
- Runner-side filtering in agent-runner removes denied/non-allowed tools after assembly
- Semantics: deny-only = blocklist, allow-only = allowlist, both = allow takes precedence (least privilege)

## Problem
`toolPolicy` on an Ensemble persona had no effect — the `AgentConfigToolPolicy` was never copied to `AgentRun.Spec.ToolPolicy`, so the env vars were never set and the runner never filtered.

Relates to #288 (Fix B: propagate + filter toolPolicy for built-in tools).

## Changes
1. **cmd/agent-runner/main.go** — reads `TOOL_POLICY_ALLOW` and `TOOL_POLICY_DENY` env vars, filters all tool types (built-in, MCP, sidecar) after assembly
2. **internal/controller/agentrun_controller.go** — serializes `ToolPolicy.Allow` and `ToolPolicy.Deny` to env vars; applied at pod build time
3. **internal/controller/ensemble_controller.go** — `lookupToolPolicyForAgent` resolves Ensemble → AgentConfig → ToolPolicySpec via `sympozium.ai/ensemble` + `sympozium.ai/agent-config` labels. Wired at stimulus AgentRun creation.
4. **internal/controller/agentrun_controller.go** — wired at sequential/pipeline AgentRun creation
5. **internal/controller/sympoziumschedule_controller.go** — wired at scheduled AgentRun creation
6. **internal/orchestrator/spawner.go** — wired at sub-agent delegation

## Verified in production
Tested with `execute_command` in the deny list on the sd-agents-collector Ensemble — confirmed tool is filtered at runner startup: `tool policy: denied tool "execute_command"`.

## Test plan
- [x] `go build ./...` — clean
- [x] Production verification — deny list filtering confirmed in collector agent logs

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -5,6 +5,7 @@ on:
     branches: [main]
   pull_request:
     branches: [main]
+  workflow_dispatch:
 
 env:
   REGISTRY: ghcr.io/velatir/sympozium
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -7,7 +7,7 @@ on:
     branches: [main]
 
 env:
-  REGISTRY: ghcr.io/sympozium-ai/sympozium
+  REGISTRY: ghcr.io/velatir/sympozium
 
 permissions:
   contents: read
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -46,8 +46,8 @@ jobs:
         run: |
           mkdir -p _site/charts .helm-packages
           # Download chart packages from all GitHub releases
-          for tag in $(gh release list -R sympozium-ai/sympozium -L 100 --json tagName -q '.[].tagName'); do
-            gh release download "$tag" -R sympozium-ai/sympozium -p 'sympozium-*.tgz' -D .helm-packages 2>/dev/null || true
+          for tag in $(gh release list -R velatir/sympozium -L 100 --json tagName -q '.[].tagName'); do
+            gh release download "$tag" -R velatir/sympozium -p 'sympozium-*.tgz' -D .helm-packages 2>/dev/null || true
           done
           # Generate index.yaml and serve chart packages from Pages
           if ls .helm-packages/*.tgz 1>/dev/null 2>&1; then
--- a/.github/workflows/integration-kind.yaml
+++ b/.github/workflows/integration-kind.yaml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 90
     env:
-      REGISTRY: ghcr.io/sympozium-ai/sympozium
+      REGISTRY: ghcr.io/velatir/sympozium
       TAG: latest
       SYMPOZIUM_NAMESPACE: sympozium-system
       TEST_NAMESPACE: default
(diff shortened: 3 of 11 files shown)


AlexsJones (changes requested):
Thanks @KSiig — this closes a real gap, and the wiring is thorough. I verified the two claims that matter:

- **The propagation hole was real**: `AgentConfigToolPolicy` was never copied onto `AgentRun.Spec.ToolPolicy` at any of the four creation sites (stimulus, sequential successor, scheduled, sub-agent spawn), so the runner env vars never existed. All four sites are now wired.
- **The filter really does cover the full tool set**: in `cmd/agent-runner/main.go` every registration point (`defaultTools()` L259, MCP L262, sidecar L297, memory L373, workflow-memory L423) precedes the new filter block, so nothing bypasses the policy — including `delegate_to_persona`, which makes this a usable workaround for #288 until advertisement is edge-aware.

Two concrete asks before this merges:

1. **Add a small table-driven unit test for the filter semantics** (deny-only, allow-only, both lists, a name in both, whitespace/empty entries). This is a security-relevant filter and the allow+deny interaction is subtle enough that it deserves a lock — `go build` and one production log line won't catch a future regression in precedence.
2. **Fix the precedence wording.** The comment (and PR body) say "both = allow takes precedence", but the code checks `denied[t.Name]` *first*, so a tool present in both lists is **denied**. That's the right least-privilege behavior — the intersection of allowed-and-not-denied — so please keep the code and correct the comment to say deny wins on conflict.

Non-blocking nit: `lookupToolPolicyForAgent` (ensemble_controller.go) and `resolveToolPolicyFromAgent` (spawner.go) are near-identical copies. If there's a sensible shared home for one exported helper, consolidating would keep the Ensemble→AgentConfig→ToolPolicySpec resolution from drifting between controller and orchestrator.

AlexsJones (approved):
Delta re-review of the two commits since my last pass (`4dd04430`, `1a2e1c84`) — both asks addressed, and the nit too:

1. **Filter semantics test** — `tool_policy_test.go` adds a table-driven suite over the extracted `applyToolPolicy` helper: deny-only, allow-only, allow+deny conflict, same-name-in-both, whitespace/empty entries, and the empty-string edge cases. This is exactly the lock the security-relevant filter needed.
2. **Precedence wording** — the comment now reads "deny wins on conflict — a tool in both lists is denied (least privilege)", matching what the code does. Code behavior unchanged, which is the right call.
3. **Consolidation (non-blocking nit, done anyway)** — `lookupToolPolicyForAgent`/`resolveToolPolicyFromAgent` are now one exported `toolpolicy.ForAgent` in a new `internal/toolpolicy` package, used at all four AgentRun creation sites, so the Ensemble→AgentConfig→ToolPolicySpec resolution can't drift.

CI is green (format/vet/build/test/codegen all pass) and the branch is mergeable. Approving — recommending to @AlexsJones for final look and merge.

PR #293, opened 2026-07-14
ci: cut build minutes — cross-compile mcp-bridge, gate fleet builds, scope caches, one verify job

AlexsJones:
Follow-up to the Actions-minutes analysis: this repo consumed ~3,900 runner-minutes in July 1–14 (~2,900 on July 8–9 alone), with the `build-and-push` fleet at ~2,200. Three changes:

## 1. Cross-compile the last QEMU straggler
`mcp-bridge` was the only Dockerfile still compiling Go **inside** the target platform — its `linux/arm64` leg ran the compiler under QEMU emulation (8.6 min avg per run, the most expensive image in the fleet). It now uses the identical `--platform=$BUILDPLATFORM` + `GOARCH=$TARGETARCH` pattern as its 17 siblings. Verified locally: `buildx build --platform linux/arm64 --target builder` produces a `GOARCH=arm64` binary in ~30s with zero emulation.

## 2. Stop paying for the fleet when nothing image-shaped changed
- `dorny/paths-filter` gates `build-and-push` on `**.go`, `go.mod/sum`, `images/**`, `web/**`, and the workflow itself — docs/chart-only pushes no longer rebuild 18 images.
- `concurrency: cancel-in-progress` per ref: back-to-back main pushes (the July 8–9 pattern) supersede in-flight runs instead of stacking full fleet builds. Safe because the newer run rewrites every `sha` + `latest` tag the cancelled one would have.

## 3. One verify job instead of five
fmt / vet / build / test / codegen were five runners, each paying checkout + setup-go + module download (~1,100 min across 82 runs this month). Now sequential steps of a single `Verify` job on one warm cache — failures still name themselves per step.

## Bonus: per-image buildx cache scopes (build + release workflows)
All 18 matrix jobs wrote `type=gha,mode=max` into the **same default cache scope**, evicting each other constantly — effectively every build was cold. Scoping by `matrix.image` gives each image a stable cache; warm rebuilds of unchanged images should drop to ~1 min.

Estimated combined effect at July's push rate: **~60–70% fewer minutes**.

Notes:
- `workflow_dispatch` and the `ghcr.io/velatir/sympozium` registry from current main are preserved untouched.
- No branch protection exists on `main`, so the job renames (`Format`/`Vet`/… → `Verify`) don't orphan any required checks.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -14,9 +14,44 @@ permissions:
   contents: read
   packages: write
 
+# A newer push supersedes an in-flight run of the same ref. The fleet is
+# tagged by sha + latest, so the superseding run rewrites every tag the
+# cancelled one would have written — back-to-back pushes stop paying for
+# the whole 18-image fleet twice.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
-  fmt:
-    name: Format
+  # Gate the expensive fleet build on whether anything that reaches an
+  # image actually changed. Verification is one cheap job and always runs.
+  changes:
+    name: Detect changes
+    runs-on: ubuntu-latest
+    if: github.event_name != 'workflow_dispatch'
+    outputs:
+      code: ${{ steps.filter.outputs.code }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: dorny/paths-filter@v3
+        id: filter
+        with:
+          filters: |
+            code:
+              - '**.go'
+              - 'go.mod'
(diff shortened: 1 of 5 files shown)


AlexsJones:
Added a second commit: PR #289 accidentally upstreamed its fork's CI plumbing (registries → `ghcr.io/velatir/sympozium` in build/release/integration-kind, docs.yml pulling chart packages from `velatir/sympozium` releases). `GITHUB_TOKEN` can't push to another org's ghcr namespace, so the next main push would have 403'd the entire image fleet. All four workflows point back at `sympozium-ai`.

PR #294, opened 2026-07-14
feat: helper to show active agent logs

kristof-ringleff:
This script has been helpful for me understanding what the currently active agents are doing (since pods are deleted after finishing with no log retention).

image

Either run it standalone or from the Makefile with `make tail-agents`. It has some quality of life features to trim off an ensemble name from the log entry to shorten output like so `make tail-agents TAIL_ARGS="--trim ensemble-team-name-"` and different agents colour differentiated. Most often you would want the skills output visible by either using `make tail-agents TAIL_ARGS="--skills"` or `make tail-agents TAIL_ARGS="--all"` (includes output from every container including ipc-bridge).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ IMAGES = controller apiserver ipc-bridge webhook agent-runner web-proxy node-pro
 		 skill-k8s-ops skill-sre-observability skill-github-gitops skill-llmfit skill-memory \
 		 llmfit-daemon mcp-bridge
 
-.PHONY: all build test clean generate manifests docker-build docker-push install help web-build web-dev web-dev-serve web-clean web-install setup-hooks integration-tests ux-tests
+.PHONY: all build test clean generate manifests docker-build docker-push install help web-build web-dev web-dev-serve web-clean web-install setup-hooks integration-tests ux-tests tail-agents
 
 all: build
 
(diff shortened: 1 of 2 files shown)


AlexsJones (approved):
Useful, and the reasoning in the cleanup trap is the good kind of comment — it explains why `kill 0` is wrong here rather than just what the code does. I verified the selector: `sympozium.ai/component=agent-run` matches what the controller actually sets (agentrun_controller.go:2253, agentrun_sandbox.go:387 and :485, spawner.go:146). `AGENT_NAMESPACE ?= default` lines up with the samples and is correctly kept distinct from `SYMPOZIUM_NAMESPACE`.

Note for anyone reading the green check: `Verify` is Go-only and there's no shellcheck in CI, so nothing here actually exercises this script. That's not a blocker for `hack/`, just worth being clear about.

Nits:

- Invoke it as `./hack/tail-agents.sh` — every other hack script in the Makefile uses the `./` prefix (lines 97-98). Works either way, just consistency.
- `SEEN` and `KIDS` grow for the life of the process. Bounded by pod count so it's a slow leak at worst, but a long-running tail on a busy namespace will accumulate.
- POLL=3 means a run that starts and finishes inside the poll window is missed entirely. Given the script's whole premise is that the logs are gone forever once the pod is deleted, that limitation deserves a line in the usage text — otherwise the first time someone misses a fast run they'll assume the script is broken rather than that they lost the race.

PR #295, opened 2026-07-15
fix(agent-runner): gate delegate_to_persona on delegation edges (Fix A, #288)

### LINKED ISSUE #288: `delegate_to_persona` is advertised to personas that have no delegation edges, and `toolPolicy` cannot remove it
## Summary

The agent-runner unconditionally registers `delegate_to_persona` in every agent's tool list, including for personas that have **zero outgoing delegation/sequential edges** in the Ensemble. The spawner correctly rejects the resulting call, but only *after* the model has already spent a tool iteration on it:

```
Delegation to "worker" failed: spawn failed: delegation denied: no delegation or
sequential relationship from "planner" to "worker" in Ensemble "dev-team"
- ensure this is correct in the ensemble definition
```

Enforcement (spawner, at call time) and advertisement (runner, at prompt-build time) disagree about who may delegate. They should agree.

## Detail

**1. The controller already knows the answer.** `injectRelationshipContext` filters the ensemble's relationships down to those where `rel.Source == personaName` (`internal/controller/agentrun_controller.go:2823-2833`), and if none remain it returns without setting `ENSEMBLE_RELATIONSHIPS` at all (`:2834`). For a persona with no outgoing edges, the controller has explicitly concluded "this persona delegates to nobody."

**2. The runner ignores that conclusion when building the tool list.** `buildRelationshipContext` correctly emits no "## Team Relationships" prompt section when the env var is absent (`cmd/agent-runner/main.go:747-763`) — but tool registration is a separate, unconditional path:

```go
// cmd/agent-runner/main.go:271-272
if toolsEnabled {
    tools = defaultTools()   // always includes ToolDelegateToPersona
```

`defaultTools()` (`cmd/agent-runner/tools.go:79`, entry at `:248`) always includes `delegate_to_persona`, and there is a test asserting this is intentional:

```go
// cmd/agent-runner/main_test.go:1158
t.Error("delegate_to_persona tool should always be registered")
```

So the model is handed a tool whose description says *"Delegate a task to another persona in your team (Ensemble)"* while having been told nothing about any team members — and no valid target exists. The tool schema is an invitation; some models take it.

**3. `toolPolicy` cannot be used to work around this.** `agentConfigs[].toolPolicy.allow/deny` on an Ensemble is never propagated to the agent pod. Grepping the tree, `ToolPolicy` is read in exactly two places outside the generated deepcopy code:

- `cmd/sympozium/main.go:10829` — to *print* an `allow:N deny:N` summary.
- `internal/webhook/policy_enforcer.go:429` — `validateToolPolicy`, which validates `AgentRun.spec.toolPolicy` against a `SympoziumPolicy`.

Nothing copies `Ensemble.spec.agentConfigs[].toolPolicy` into `AgentRun.spec.toolPolicy`, and the runner never reads a tool policy when building its tool list. An operator who writes a careful allow-list omitting `delegate_to_persona` (or an explicit `deny: [delegate_to_persona]`) gets no enforcement and no warning — the field silently does nothing.

## Reproduction

1. Define an Ensemble with personas `planner` and `worker` and exactly one delegation edge, `dev-orchestrator → worker`. Give `planner` a `toolPolicy.allow` list that does not include `delegate_to_persona`.
2. Put anything in the planner's context that mentions delegating to `worker`. (In our case the planner and the orchestrator share a SkillPack; since skills are selected per-*pack*, not per-skill, the planner mounts the orchestrator's `plan-orchestrate` skill, which instructs the reader to call `delegate_to_persona` with target `worker`.)
3. Run the planner. It calls `delegate_to_persona(targetPersona: "worker")`.
4. The spawner denies it at `internal/orchestrator/spawner.go:279-282`.

## Impact

The guard holds — no unauthorised run is spawned, so this is not a security issue. The costs are:

- **Wasted iterations.** Each attempt burns a tool call against `MAX_TOOL_ITERATIONS` and an LLM round-trip. A model that retries burns several.
- **Misleading error.** *"ensure this is correct in the ensemble definition"* points the operator at a correct definition. The runner offering the tool is what's incorrect.
- **`toolPolicy` is a trap.** It reads like an enforcement boundary and is documented as "defines which tools this agent is allowed to use" (`api/v1alpha1/ensemble_types.go:214`), but has no effect on built-in tools. Anyone relying on it for `deny: [write_file]` etc. has a false sense of containment.

## Suggested fixes

Either would resolve the reported error; they're independent and both look worthwhile.

**A. Only register `delegate_to_persona` when the persona has at least one delegation/sequential edge.** The runner already has the signal — `ENSEMBLE_RELATIONSHIPS` is absent or contains no `delegation`/`sequential` entries. Gate the tool on the same condition that gates the prompt section, so advertisement matches what the spawner will actually authorise. `defaultTools()` would take the relationship set (or a bool) as a parameter; `main_test.go:1158` would be updated to assert registration *when a delegation edge exists* and absence otherwise.

Optionally, when the tool *is* registered, enumerate the valid targets in its description — the spawner will reject anything else anyway, so naming them can only reduce wasted calls.

**B. Make `toolPolicy` real, or make it loudly not-real.** Propagate `Ensemble.spec.agentConfigs[].toolPolicy` into `AgentRun.spec.toolPolicy`, plumb it to the pod, and filter `defaultTools()` through it (the MCP path already has exactly this logic in `internal/mcpbridge/manifest.go:141`, `filterTools(tools, allow, deny)` — same allow-then-deny semantics would apply). If that's not the intended direction, the field should be rejected by the webhook or at minimum documented as MCP-only, because today it silently does nothing for built-in tools.

If **A** lands, the specific error above goes away. **B** is the more general fix: without it, there is no supported way for an operator to take a built-in tool away from a persona.

EvilGorilla13:
## Summary

Fixes #288 (Fix A): `delegate_to_persona` was unconditionally advertised to every agent persona even when no delegation edges existed in the Ensemble. This PR gates its registration on `ENSEMBLE_RELATIONSHIPS` containing at least one `delegation` or `sequential` edge — mirroring the existing `SUBAGENTS_ENABLED` pattern used for `spawn_subagents`.

**Bonus**: the tool schema's `targetPersona` field now includes an `"enum"` of valid targets, so the LLM sees exactly who it can delegate to rather than an open-ended string.

## Changes

**`cmd/agent-runner/tools.go`**
- Remove `ToolDelegateToPersona` from the unconditional base tool slice
- Add conditional registration block gated on `delegationTargets()` returning at least one target
- Add `delegationTargets()` helper: parses `ENSEMBLE_RELATIONSHIPS`, returns `[]string` of persona names reachable via `delegation` or `sequential` edges; returns `nil` (not registered) when the env var is unset, empty, or contains only `supervision` edges

**`cmd/agent-runner/main_test.go`**
- Update `TestDefaultTools_AlwaysIncludesCoreTool` — remove the assertion that `delegate_to_persona` is always present (it no longer is)
- Add 5 new tests:
  - `TestDefaultTools_DelegateAbsentWithoutEdges` — not registered when `ENSEMBLE_RELATIONSHIPS` is unset
  - `TestDefaultTools_DelegateAbsentWithSupervisionEdgeOnly` — not registered for supervision-only edges
  - `TestDefaultTools_DelegateRegisteredWithDelegationEdge` — registered when a delegation edge is present
  - `TestDefaultTools_DelegateRegisteredWithSequentialEdge` — registered when a sequential edge is present
  - `TestDefaultTools_DelegateEnumeratesValidTargets` — `"enum"` field contains exactly the reachable persona names

## Test plan

- [ ] `go test -race ./cmd/agent-runner/...` passes (all 5 new tests + updated existing test)
- [ ] `go vet ./...` clean
- [ ] `gofmt` produces no diff
- [ ] Manually verify: an agent with no outgoing edges does not see `delegate_to_persona` in its tool list
- [ ] Manually verify: an agent with a `delegation` edge sees `delegate_to_persona` with correct `enum` targets

### DIFF
--- a/cmd/agent-runner/main_test.go
+++ b/cmd/agent-runner/main_test.go
@@ -1142,20 +1142,91 @@ func TestDefaultTools_SubagentsEnabledWhenSet(t *testing.T) {
 
 func TestDefaultTools_AlwaysIncludesCoreTool(t *testing.T) {
 	tools := defaultTools()
-	var hasExec, hasDelegate bool
+	var hasExec bool
 	for _, tool := range tools {
 		if tool.Name == ToolExecuteCommand {
 			hasExec = true
-		}
-		if tool.Name == ToolDelegateToPersona {
-			hasDelegate = true
 		}
 	}
 	if !hasExec {
 		t.Error("execute_command tool should always be registered")
 	}
-	if !hasDelegate {
-		t.Error("delegate_to_persona tool should always be registered")
+}
+
+func TestDefaultTools_DelegateAbsentWithoutEdges(t *testing.T) {
+	t.Setenv("ENSEMBLE_RELATIONSHIPS", "")
+	tools := defaultTools()
+	for _, tool := range tools {
+		if tool.Name == ToolDelegateToPersona {
+			t.Error("delegate_to_persona tool should not be registered when ENSEMBLE_RELATIONSHIPS is not set")
+		}
+	}
+}
+
+func TestDefaultTools_DelegateAbsentWithSupervisionEdgeOnly(t *testing.T) {
+	t.Setenv("ENSEMBLE_RELATIONSHIPS", `[{"target":"supervisor","type":"supervision"}]`)
+	tools := defaultTools()
+	for _, tool := range tools {
+		if tool.Name == ToolDelegateToPersona {
+			t.Error("delegate_to_persona tool should not be registered for supervision-only edges")
+		}
+	}
+}
+
+func TestDefaultTools_DelegateRegisteredWithDelegationEdge(t *testing.T) {
+	t.Setenv("ENSEMBLE_RELATIONSHIPS", `[{"target":"writer","type":"delegation"}]`)
+	tools := defaultTools()
+	var found bool
+	for _, tool := range tools {
+		if tool.Name == ToolDelegateToPersona {
+			found = true
+			break
+		}
+	}
+	if !found {
+		t.Error("delegate_to_persona tool should be registered when a delegation edge is present")
+	}
+}
+
+func TestDefaultTools_DelegateRegisteredWithSequentialEdge(t *testing.T) {
+	t.Setenv("ENSEMBLE_RELATIONSHIPS", `[{"target":"reviewer","type":"sequential"}]`)
+	tools := defaultTools()
+	var found bool
+	for _, tool := range tools {
+		if tool.Name == ToolDelegateToPersona {
+			found = true
+			break
+		}
+	}
+	if !found {
+		t.Error("delegate_to_persona tool should be registered when a sequential edge is present")
+	}
+}
+
+func TestDefaultTools_DelegateEnumeratesValidTargets(t *testing.T) {
+	t.Setenv("ENSEMBLE_RELATIONSHIPS", `[{"target":"writer","type":"delegation"},{"target":"reviewer","type":"sequential"}]`)
+	tools := defaultTools()
+	var delegateTool *ToolDef
+	for i := range tools {
+		if tools[i].Name == ToolDelegateToPersona {
+			delegateTool = &tools[i]
+			break
+		}
+	}
+	if delegateTool == nil {
+		t.Fatal("delegate_to_persona tool not registered")
+	}
+	props, _ := delegateTool.Parameters["properties"].(map[string]any)
+	targetParam, _ := props["targetPersona"].(map[string]any)
+	enum, _ := targetParam["enum"].([]string)
+	if len(enum) != 2 {
+		t.Fatalf("expected 2 enum values, got %d: %v", len(enum), enum)
+	}
+	wantTargets := map[string]bool{"writer": true, "reviewer": true}
+	for _, v := range enum {
+		if !wantTargets[v] {
+			t.Errorf("unexpected enum value %q", v)
+		}
 	}
 }
 
--- a/cmd/agent-runner/tools.go
+++ b/cmd/agent-runner/tools.go
@@ -244,19 +244,24 @@ func defaultTools() []ToolDef {
 				"required": []string{"name", "action"},
 			},
 		},
-		{
+	}
+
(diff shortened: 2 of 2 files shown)


AlexsJones (changes requested):
Fix A is the right shape and this is close, but a few things to sort first.

Worth noting up front: no CI has run on this. The workflow is sitting in `action_required` pending approval, so nothing here has been verified by the build — including the gofmt point below.

What's right: the `delegation || sequential` predicate exactly mirrors the spawner's check at `spawner.go:277-278`, which is the property that matters — advertisement now agrees with enforcement. The env var shape matches what `injectRelationshipContext` marshals (`agentrun_controller.go:3223-3227`). And reading the env inside `defaultTools()` rather than threading it through as a parameter is better than what I suggested in the issue — it matches the `SUBAGENTS_ENABLED` pattern at `tools.go:273`.

Changes:

1. **gofmt.** There's a second consecutive blank line before `delegationTargets`. `Verify` runs `gofmt -l` (`build.yaml:66`) and will fail on it. Run `gofmt -s -w .`.

2. **Dedup the targets.** An ensemble with both a `delegation` and a `sequential` edge from the same source to the same target produces `enum: ["worker", "worker"]`. Cheap to fix, and duplicate enum values are the kind of thing that makes a model second-guess itself.

3. **The sequential case reintroduces a small version of the bug this fixes.** `buildRelationshipContext` only lists `delegation` edges in the prompt — `case "delegation"` in `main.go`. So a persona with only sequential edges now gets `delegate_to_persona` registered with a populated `enum`, but `buildRelationshipContext` returns `""` and it gets no "## Team Relationships" section at all. The tool says "you may delegate to reviewer", the prompt says nothing about a reviewer.

   The spawner does authorize sequential, so gating the tool on both types is correct — it's the prompt path that's lagging. I'd rather fix `buildRelationshipContext` to describe sequential edges than narrow the gate. Happy for that to be a follow-up if you'd prefer to keep this PR tight, but let's decide rather than leave it implicit.

Not asking you to touch `toolPolicy` — #289 covers that half.

PR #296, opened 2026-07-15
feat: allow ranged reads via read_file native tool

kristof-ringleff:
Agents with only the read_file tool available would not be able to process more than the 8000 bytes capped by the tool at the moment. This PR expands the tool to allow ranged reads via offset and limit.

This should hopefully also remove some ipc execute_command round trips for calls like head or sed for specific sections of a file.

### DIFF
--- a/cmd/agent-runner/tools.go
+++ b/cmd/agent-runner/tools.go
@@ -112,14 +112,24 @@ func defaultTools() []ToolDef {
 			},
 		},
 		{
-			Name:        ToolReadFile,
-			Description: "Read the contents of a file from the pod filesystem. Paths under /workspace, /skills, /tmp, and /ipc are accessible.",
+			Name: ToolReadFile,
+			Description: "Read the contents of a file from the pod filesystem. Paths under /workspace, /skills, /tmp, and /ipc are accessible. " +
+				"Responses are capped at 8000 bytes; for larger files use 'offset' and 'limit' to read a specific line range, " +
+				"following the continuation hint appended to truncated output.",
 			Parameters: map[string]any{
 				"type": "object",
 				"properties": map[string]any{
 					"path": map[string]any{
 						"type":        "string",
 						"description": "Absolute path to the file to read.",
+					},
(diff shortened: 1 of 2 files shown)


AlexsJones (approved):
Reviewed the full diff. This closes a real capability gap — an agent with only `read_file` was hard-stopped at 8000 bytes with no way to see the rest of the file — and it does it in a way that composes well with how models actually use tools:

- **The continuation hint is exact.** I checked the arithmetic: with 41-byte lines the cap admits 195 complete lines (7995 bytes), and the hint says `truncated at line 195 of 500; continue with offset=196` — and `TestReadFileTool_CapTruncatesAtLineBoundary` doesn't just assert the hint, it *follows* it and verifies the continuation read resumes at line 196 with no gap or overlap. That round-trip test is exactly the right thing to pin.
- **Edge cases are handled**: cut always lands on a complete line so line counting stays truthful; a single line over the cap gets an honest `truncated mid-line` message instead of a bogus continuation offset; offset past EOF errors with the real line count; the trailing-newline line isn't counted as a phantom line.
- **The fast path is untouched**: `offset==1 && limit==0 && len<=8000` returns the file byte-for-byte as before, so existing behavior for small files is unchanged.
- The tool description teaches the model the pattern (cap → follow the hint), and `TestReadFileToolDefAdvertisesRange` locks the schema with `path` still the only required field.

Two non-blocking notes:

1. The whole file is still read into memory (`os.ReadFile`) before ranging — pre-existing behavior, but ranged reads will invite use on exactly the large files where it stings. A follow-up could seek/scan instead; not needed for this PR.
2. A line that exceeds the cap is unreachable in full by any `offset` (inherent to line-based ranges). If that ever bites in practice, a byte-offset escape hatch would be the fix — again follow-up territory.

CI green. Approving — recommending to @AlexsJones for final look and merge. (Heads-up for merge ordering only: #238 also touches `cmd/agent-runner/tools.go`; whichever lands second may need a trivial rebase.)

PR #303, opened 2026-07-20
feat: Add edit_file tool with multi edit support

kristof-ringleff:
This PR introduces a new native tool to edit files to avoid needing to supply the whole file for updates using the write_tool. Additionally the native read and write tools have been aligned with the edit tool around path resolution and symlinks escaping the path restrictions.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -194,13 +194,14 @@ Add new tests to the `test-integration` target in the `Makefile`.
 
 ## Agent Tools
 
-The agent-runner has 7 built-in tools defined in `cmd/agent-runner/tools.go`:
+The agent-runner has 8 built-in tools defined in `cmd/agent-runner/tools.go`:
 
 | Tool | Category | Description |
 |------|----------|-------------|
 | `execute_command` | IPC (sidecar) | Run shell commands in the skill sidecar |
 | `read_file` | Native | Read file contents |
 | `write_file` | Native | Write/create files |
+| `edit_file` | Native | Apply one or more exact-string (unique-match) replacements to a file (atomic, all-or-nothing) |
 | `list_directory` | Native | List directory contents |
 | `send_channel_message` | IPC (bridge) | Send messages to Telegram/Slack/Discord/WhatsApp |
 | `fetch_url` | Native | HTTP GET a URL and return the body |
--- a/cmd/agent-runner/edit_tool.go
+++ b/cmd/agent-runner/edit_tool.go
@@ -1,0 +1,281 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// --- edit_file native tool (runs in the agent container) ---
+//
+// Applies an ordered, all-or-nothing batch of exact-string replacements. Each
+// old_string must match the file's current bytes exactly once (unless
+// replace_all is set); zero or ambiguous matches fail loudly with line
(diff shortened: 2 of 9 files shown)


AlexsJones (approved):
Thanks @kristof-ringleff — this is a strong contribution. I went through the edit semantics, the sandbox/path story, policy integration, and the tests. Approving; details below.

## What's good

**Edit semantics** (`cmd/agent-runner/edit_tool.go`)
- Exact-string, unique-match-or-`replace_all` replacement with all-or-nothing batches is the right design, and the invariants are actually enforced: `applyEdits` rejects empty `old_string`, identical old/new, zero matches, and ambiguous matches — with line numbers in the error (`matches %d times at lines %s`), which is exactly the feedback a model needs to self-correct. Sequential application (edit *i* sees the result of *0..i-1*) is documented in the schema description and locked in by `TestApplyEdits_Chaining` / `TestEditFileTool_Sequential`.
- The atomic temp-file + fsync + rename write (`atomicWrite`) with mode preservation is careful work; `TestEditFileTool_PreservesMode` and the failure-path tests verify the file on disk is untouched on any error, not just that an error string came back.
- Editing *through* a symlink operates on the resolved target so the rename doesn't clobber the link itself — with a test for precisely that (`TestEditFileTool_EditsThroughSymlink`). Nice catch.
- Go string ops are byte-exact, so non-UTF-8 content can't be silently corrupted — a non-matching `old_string` just fails loudly. Missing file → `resolvePath`'s "cannot access" error rather than a create, matching the tool's contract.

**Path safety** (`cmd/agent-runner/paths.go`)
- Factoring out `resolvePath`/`resolveWritableTarget` and retrofitting them onto `read_file`/`write_file` fixes two real pre-existing holes on main: the old `strings.HasPrefix` check counted `/tmpfoo` as under `/tmp` (`underAnyPrefix` now compares on segment boundaries, covered by `TestWriteFileTool_PrefixBoundary`), and a symlink placed under an allowed root could previously read or write outside it. The deepest-existing-ancestor resolution handles the not-yet-existing-file case, and resolving the roots themselves avoids false denials on mount-point-symlinked roots. The escape tests assert the outside file was *not* touched, not just that a denial string appeared.

**Policy integration** — `edit_file` is registered in `defaultTools()` so it flows through `applyToolPolicy`'s name-based allow/deny like every other tool, and the `builtinToolNames` map in `internal/webhook/policy_enforcer.go` is updated per that file's sync contract. Docs updated in AGENTS.md, docs/concepts/skills.md, docs/getting-started.md.

**Schema quality** — the `edit_file` description in `tools.go` is genuinely good LLM guidance (uniqueness rule, `replace_all` escape hatch, "read the file first", empty-string-deletes, all-or-nothing), and `TestEditToolDefs` pins the required fields.

## Minor notes (none blocking)

1. `os.MkdirTemp(".", "outside-roots")` in `tools_test.go` litters the package dir if a test hard-crashes; `outside-roots*` in `.gitignore` would be belt-and-braces.
2. A non-string `old_string` silently coerces to `""` and reports "old_string is required" — a "must be a string" branch would be marginally kinder to the model.
3. Allow-list-mode deployments won't pick this up automatically — operators must add `edit_file` explicitly; worth a release-note line.

## Cross-PR coordination

#302 (open) overlaps on `docs/concepts/skills.md`; no functional duplication (#302 adds no tools), but whichever merges second will need a trivial rebase on the skills.md tools table.

CI is green (Detect changes + Verify SUCCESS; build-and-push skipped as expected). Recommending to @AlexsJones for merge.

PR #304, opened 2026-07-20
fix(controller+agent-runner): wire sidecar-driven task mode end-to-end

KSiig:
## Summary

Wires the sidecar-driven task mode (object-form `spec.task`) end-to-end. PR #302 shipped the polymorphic TaskSpec type, taskmodes registry, and SidecarDrivenHandler but never wired the dispatcher into the AgentRun reconciler and the agent-runner had no `AGENT_MODE=prompt-server` branch. Result: pods with object-form task exited 1 in ~5s with `FATAL: TASK env var is empty`.

## Changes

**Controller (`internal/controller/agentrun_controller.go`)**

- `TASK` env only set when task is string-form OR has a prompt. Object-form tasks produce `TASK=''` which trips the agent-runner's empty-TASK guard.
- `USE_CONTEXT` env propagated from `AgentRun.Spec.UseContext` (default true on nil) per the `internal/ipc` protocol doc.
- `applyTaskModeToAgentContainer` called on object-form tasks; sets `AGENT_MODE=prompt-server` for sidecar-driven mode.
- `resolveTaskModeAdjustments` called in BOTH execution paths (Job via `reconcilePending`, and `reconcilePendingAgentSandbox`). Unknown modes + handler validation errors surface on `AgentRun.status.error` instead of producing a confusing 5s exit.
- `buildContainers` accepts sidecar adjustments; in the sidecar loop, `OverrideCommand` replaces `sc.sidecar.Command` and `AddEnv` is appended after `SYMPOZIUM_SKILL_PACK`.

**Agent-runner (`cmd/agent-runner/`)**

- `main.go`: `AGENT_MODE=prompt-server` branch early-returns into `runPromptServerAgent`. Skips the main agent loop entirely.
- `prompt_server.go` (NEW): watches `/ipc/prompts/`, `/ipc/context/`, `/ipc/done` via polling. Builds ONE provider per run, serialised via `sync.Mutex`. On `/ipc/done`: copies `/ipc/run-result.json` to `/ipc/output/result.json` (sidecar owns the lifecycle) and emits the `__SYMPOZIUM_RESULT__` marker the controller parses to transition the AgentRun.
- `provider.go`: `LLMProvider` interface gains `Prompt(ctx, prompt, useContext, schema)` and `ResetContext()`.
- `provider_{anthropic,openai,bedrock}.go`: implement `Prompt` + `ResetContext`. Anthropic/Bedrock use json.Valid post-hoc; OpenAI forwards schema as `response_format.json_schema` per protocol doc.

## Tests

- 7 new `buildContainers` tests (object-task sets AGENT_MODE, sets USE_CONTEXT, omits TASK; string-task still sets TASK; sidecar adjustments override command and add env).
- 3 new `prompt_server` tests (request → result, clear-context handled, useContext propagated).
- `mockProvider` in `provider_test.go` updated to satisfy the extended `LLMProvider` interface.
- Full suite: 627 tests pass across the controller + agent-runner packages.

## Why this isn't a port of the old prompt-server work

A previous attempt added a post-loop prompt service gated by `ENABLE_PROMPT_SERVICE=true`. That shape is incompatible with the new object-form `spec.task` — this PR implements an `AGENT_MODE=prompt-server` early-return instead, which is the dispatcher-driven shape the taskmodes registry expects. The per-provider `Prompt()` and `ResetContext()` semantics (tool-use suppression, schema forwarding) are ported verbatim because they didn't change between the two shapes.

## Test plan

```bash
go build ./...
go test -count=1 -timeout=180s ./...
```


AlexsJones (changes requested):
The wiring fix at the heart of this PR (e27cdaf0) is genuinely good work — the `resolveTaskModeAdjustments` dispatch surfacing handler validation on the CR's `status.error` instead of a cryptic 5-second pod death, the TASK-env gate for object-form tasks, the prompt-server entry point with run-timeout respect and validate-before-publish on `run-result.json`, and real tests on both the controller and agent-runner sides. It also confirms something important about #302: the type/registry work alone isn't functional end-to-end, and this is the missing half.

The problem is the packaging: the PR branch is the velatir fork's `main` diffed against upstream, so alongside the wiring it carries things that can't land here as-is:

- **All of still-open #302** (78fe99e) — that PR is already approved upstream and should merge on its own; this one should stack on it.
- **`internal/apiserver/pricing_handlers_test.go` adds `t.Skip("…not yet wired on the velatir fork; see velatir/sympozium#10")` to eight tests.** Upstream *has* wired those routes (#291 merged last week) — merging this would silently disable a live upstream test suite with a comment pointing at a fork-internal issue.
- **An unrelated CI overhaul of `.github/workflows/build.yaml`** (concurrency groups, paths-filter gating, consolidated verify job, per-image cache scopes). It honestly looks like a nice improvement — but it deserves its own PR where it can be discussed on its merits, not riding inside a bug fix.
- **`CHANGELOG.md`, `.release-please-manifest.json`, and chart version bumps** — release-please owns those upstream and they'll conflict on every release.
- Assorted fork-side extras: `cost_estimation_design.md`, the detailed-logging feature + docs, `hack/tail-agents.sh`.

Could you rebase this onto upstream `main` so the PR contains just the wiring commit, marked as depending on #302 (or including #302's commits unchanged if you'd rather it merge as one unit — maintainer's call which order lands)? The `docs.yml` `velatir/sympozium` → `sympozium-ai/sympozium` release-download fix is a genuine upstream fix too — happy to see that as a tiny separate PR. Same for the CI overhaul and detailed logging if you want to upstream them; on a quick read they'd both be welcome.

Requesting changes only for the repackaging — the fix itself reads correct and I'd expect a fast approve on the focused version.

PR #306, opened 2026-07-21
fix(controller+agent-runner): wire sidecar-driven task mode end-to-end

KSiig:
## Summary

Wires the sidecar-driven task mode (object-form `spec.task`) end-to-end. PR #302 shipped the polymorphic TaskSpec type, taskmodes registry, and SidecarDrivenHandler but never wired the dispatcher into the AgentRun reconciler and the agent-runner had no `AGENT_MODE=prompt-server` branch. Result: pods with object-form task exited 1 in ~5s with `FATAL: TASK env var is empty`.

## Changes

**Controller (`internal/controller/agentrun_controller.go`)**

- `TASK` env only set when task is string-form OR has a prompt. Object-form tasks produce `TASK=''` which trips the agent-runner's empty-TASK guard.
- `USE_CONTEXT` env propagated from `AgentRun.Spec.UseContext` (default true on nil) per the `internal/ipc` protocol doc.
- `applyTaskModeToAgentContainer` called on object-form tasks; sets `AGENT_MODE=prompt-server` for sidecar-driven mode.
- `resolveTaskModeAdjustments` called in BOTH execution paths (Job via `reconcilePending`, and `reconcilePendingAgentSandbox`). Unknown modes + handler validation errors surface on `AgentRun.status.error` instead of producing a confusing 5s exit.
- `buildContainers` accepts sidecar adjustments; in the sidecar loop, `OverrideCommand` replaces `sc.sidecar.Command` and `AddEnv` is appended after `SYMPOZIUM_SKILL_PACK`.

**Agent-runner (`cmd/agent-runner/`)**

- `main.go`: `AGENT_MODE=prompt-server` branch early-returns into `runPromptServerAgent`. Skips the main agent loop entirely.
- `prompt_server.go` (NEW): watches `/ipc/prompts/`, `/ipc/context/`, `/ipc/done` via polling. Builds ONE provider per run, serialised via `sync.Mutex`. On `/ipc/done`: copies `/ipc/run-result.json` to `/ipc/output/result.json` (sidecar owns the lifecycle) and emits the `__SYMPOZIUM_RESULT__` marker the controller parses to transition the AgentRun.
- `provider.go`: `LLMProvider` interface gains `Prompt(ctx, prompt, useContext, schema)` and `ResetContext()`.
- `provider_{anthropic,openai,bedrock}.go`: implement `Prompt` + `ResetContext`. Anthropic/Bedrock use json.Valid post-hoc; OpenAI forwards schema as `response_format.json_schema` per protocol doc.

## Tests

- 7 new `buildContainers` tests (object-task sets AGENT_MODE, sets USE_CONTEXT, omits TASK; string-task still sets TASK; sidecar adjustments override command and add env).
- 3 new `prompt_server` tests (request → result, clear-context handled, useContext propagated).
- `mockProvider` in `provider_test.go` updated to satisfy the extended `LLMProvider` interface.
- Full suite: 627 tests pass across the controller + agent-runner packages.

## Why this isn't a port of the old prompt-server work

A previous attempt added a post-loop prompt service gated by `ENABLE_PROMPT_SERVICE=true`. That shape is incompatible with the new object-form `spec.task` — this PR implements an `AGENT_MODE=prompt-server` early-return instead, which is the dispatcher-driven shape the taskmodes registry expects. The per-provider `Prompt()` and `ResetContext()` semantics (tool-use suppression, schema forwarding) are ported verbatim because they didn't change between the two shapes.

## Test plan

```bash
go build ./...
go test -count=1 -timeout=180s ./...
```


AlexsJones (changes requested):
Thanks @KSiig — the wiring commit itself (72537291) is a real step up from #304's: `USE_CONTEXT` propagation, task-mode validation errors surfacing on `status.error` instead of a 5-second crash loop, `resolveTaskModeAdjustments` called in both execution paths, and the provider-once + mutex rationale written down where the next reader needs it. The provider abstraction reads cleanly across all three backends. Requesting changes on four points — the first undoes the headline fix.

**1. The empty-TASK fatal still fires before the prompt-server branch (blocker).**
In `cmd/agent-runner/main.go` the order is:

```go
task := getEnv("TASK", "")
// … fallback to /ipc/input/task.json …
if task == "" {
	fatal("TASK env var is empty and no /ipc/input/task.json found")
}

if getEnv("AGENT_MODE", "") == "prompt-server" {
	runPromptServerAgent()
	return
}
```

By this PR's own controller change (and its `buildContainers` test asserting object-form tasks omit `TASK`), a sidecar-driven pod starts with no `TASK` env and nothing writes `/ipc/input/task.json` — so it dies on the exact `FATAL: TASK env var is empty` this PR sets out to fix, before the `AGENT_MODE` check is reached. Move the prompt-server early-return above the task-resolution block; prompt-server mode has no initial task by design (`Task: ""` in `PromptServerDeps` already says so).

**2. The branch is still the fork's main, not the focused branch asked for on #304 (blocker, carried over).**
The eight `t.Skip("pricing handler route not yet wired into the apiserver mux on the velatir fork…")` on upstream pricing tests are still in the diff — those routes ARE wired upstream since #291, so merging this would silently disable real coverage. Also still bundled: still-open #302's content, the `build.yaml` overhaul plus the `ghcr.io/velatir` registry commits (52d64f3c, 7016f2e9), release-please/CHANGELOG/chart version bumps, `detailed_logging.go` + its guide, `cost_estimation_design.md`, and `hack/tail-agents.sh`. The actual fix is one clean commit — 72537291, 19 files, +1188/−97. Cherry-picking that commit onto a branch cut from `sympozium-ai/sympozium` main (noting the #302 dependency in the body and rebasing once #302 merges) gets this to a reviewable, mergeable diff. Totally understand the agent-tooling context from your #302 comment — happy to re-review as soon as the focused branch is up; the detailed-logging and CI work would be welcome as their own PRs.

**3. The `/ipc/done` signal can be lost, leaving the pod running until the run timeout.**
In `runPromptServiceLoop`, the done-watcher goroutine does a non-blocking send on the unbuffered `doneCh` and then returns. If the main loop is inside `processRequest` (mid-LLM-call) at that moment, the send takes the `default:` branch, the goroutine exits, and nothing ever re-signals — the loop polls until `effectiveRunTimeout` kills the pod and the run is reported as timed out instead of done. `close(doneCh)` instead of the send fixes it: a closed channel satisfies the receive whenever the loop next reaches `select`.

**4. A torn request file is dropped permanently, and result writes aren't atomic despite the doc comment.**
`knownRequests[name] = true` is set before `ReadFile`/`Unmarshal`, so if the 150 ms poll catches a request file mid-write, the partial JSON is logged as "dropping invalid request" and never retried — the sidecar waits forever for its `result-{id}.json`. Mark a request known only after a successful parse (or write an error result for it so the sidecar unblocks). Symmetrically, `writePromptResult`'s comment says "atomically writes", but `os.WriteFile` straight to the final path is not atomic — write to a temp name in the same directory and `os.Rename`, so the fsnotify-watching sidecar never reads partial bytes.

Points 3 and 4 are small, mechanical fixes inside `prompt_server.go`; point 1 is a two-line move. Point 2 is the same ask as #304, but with the fix isolated in 72537291 it should be a quick rebase this time. As before: this repo runs no CI on pull requests, so the "627 tests pass" claim is verified by reading, not execution — and point 1 is exactly the kind of gap that slips past unit tests that call `buildContainers` and `runPromptServiceLoop` directly rather than going through `main()`.

PR #308, opened 2026-07-21
feat(agent-runner): log LLM I/O in prompt-service loop (VEL-1203)

KSiig:
## Summary

The sidecar-driven agent-runner (`AGENT_MODE=prompt-server`) only logged per-turn metadata — never the assembled prompt sent to the LLM or the LLM's actual response payload. When the sidecar loops on the same discovery fetches (a real failure mode observed during VEL-1174 verification on 2026-07-21), there's no way to inspect the LLM's chain-of-thought from `kubectl logs`. OpenTelemetry traces don't surface in normal triage.

This PR mirrors the regular Path A logging pattern by emitting two new lines alongside the existing `prompt service:` metadata block:

- `prompt service req: requestId=… promptBytes=… schemaBytes=… prompt=…`
- `prompt service res: requestId=… status=… payloadKind=parsed|content|error payloadBytes=… payload=…`

The `res` line prefers `result.Parsed` (the actual LLM decision the sidecar acts on) over the textual `Content` when structured output was used; falls back to `Content` otherwise; captures `Error` for the error path.

Both lines are byte-capped via `PROMPT_SERVICE_LOG_MAX_BYTES` (default 2048, set to 0 to disable entirely). The cap is applied with a `...[truncated N bytes]" marker so operators can correlate the line with the full size.

## Why this is not a privacy concern

The regular Path A flow already logs the full LLM conversation and tool calls. PR #302's prompt-server implementation simply didn't carry that pattern through. There's no sidecar-driven reason to omit prompt content — the same retention policy applies.

## Bases

Branched from PR #302's head commit (`c0bcf81`). Branch `kaspersiig/vel-1203-sidecar-driven-mode-agent-runner-prompt-server-doesnt-log` on `velatir/sympozium`.

## Dependency / merge order

This PR's diff against `sympozium-ai/sympozium:main` currently includes all of PR #302 because PR #302 is unmerged. Once PR #302 merges, this PR should be **rebased onto the new main** (drop the PR #302 commits, keep only `75de8ab`) before merge. The expected diff after rebase is just:

- `cmd/agent-runner/prompt_service.go` — 72 lines added (new env knob + helpers + 4 lines of wiring in `processRequest`)
- `cmd/agent-runner/prompt_service_logging_test.go` — new (236 lines)

I'll handle the rebase + force-push after PR #302 lands to keep the diff minimal.

## Tests

`cmd/agent-runner/prompt_service_logging_test.go` covers:

- happy path (req/res full-content emitted)
- truncation marker (req/res over-cap, "truncated N bytes" appended)
- env-var disable (`PROMPT_SERVICE_LOG_MAX_BYTES=0` suppresses both lines)
- parsed-vs-content preference on the res path
- error-path capture

`go test ./cmd/agent-runner/...` — full suite (including pre-existing tests) passes.

## Out of scope

- SkillPack or infra changes (none needed).
- Changes to the existing metadata line.

## Related

- VEL-1203 — Sidecar-driven mode (agent-runner prompt-server) does not log LLM output
- VEL-1174 — Sidecar-driven task mode is not wired end-to-end (parent feature; verification surfaced this logging gap)
- VEL-1202 — Collector fetch pipeline doesn't consume pathPatterns as fetch candidates
- PR #302 — introduced `prompt_service.go` without the full conversation logging pattern


KSiig:
Heads up: this PR currently diffs against `sympozium-ai:sympozium:main` with all of PR #302 included because PR #302 is unmerged. After PR #302 merges, I'll rebase this branch onto the new upstream main (drop PR #302's commits, keep only `75de8ab`) so the diff collapses to just the 2 files for VEL-1203. The implementation is the same regardless of when PR #302 lands — only the merge-base presentation changes.

PR #312, opened 2026-07-23
fix(crd): accept both string-form and object-form spec.task at admission

KSiig:
## Problem

PR #302 (already merged) landed polymorphic spec.task in the Go API and controller but shipped a broken CRD schema (type: object + x-kubernetes-preserve-unknown-fields) that the apiserver validates as object-only. Every internal creator that uses NewStringTask — the CLI, apiserver, channel router, schedules, spawner, webproxy, delegation successors — is rejected at admission with 'spec.task: must be of type object'. Unit tests missed it because fake clients skip OpenAPI validation.

This was called out as a blocking ask before merge on the PR #302 review (issuecomment 5033007953):

> The CRD schema rejects string-form tasks at admission. spec.task is declared as type: object with x-kubernetes-preserve-unknown-fields: true, but preserve-unknown-fields only disables pruning, it does not relax type checking. The fix is small: make the field typeless with +kubebuilder:validation:Type="" alongside the preserve-unknown-fields marker, re-run make manifests.

## Fix

Three changes ship together — same scope as the blocking review:

1. **CRD schema is typeless** — switched from hand-maintained type: object + properties to +kubebuilder:validation:Schemaless + +kubebuilder:validation:XPreserveUnknownFields. Re-running make manifests regenerates all three CRD files in lockstep (config/crd/bases/, charts/sympozium/crds/, charts/sympozium-crds/templates/) with no type field — accepts both string and object forms. Verified via kubectl --dry-run=server against the live apiserver.
2. **reconcilePending nil-task fail-fast** — second smaller ask. Was rendering an empty TASK env var and silently FATAL'ing inside the agent-runner; now r.failRun(ctx, agentRun, ...) with a clear status.error.
3. **reconcilePendingAgentSandbox dispatch error** — first smaller ask. Was bubbling the error to the reconciler (requeue forever); now matches the Job path's r.failRun pattern.

## Verification

test/integration/test-crd-task-polymorphism.sh exercises both shapes against the live apiserver:

```
## Case 1: string-form task (legacy Path A)
✓ string-form task accepted at CRD schema layer

## Case 2: object-form task (Path B / sidecar-driven)
✓ object-form task accepted at CRD schema layer

## Case 3: empty string-form task (schema accepts; controller nil-check rejects at runtime)
✓ empty string-form task accepted at CRD schema layer

All cases passed at CRD schema layer — spec.task is properly polymorphic.
```

## Diff

7 files, +282 / -87:
- api/v1alpha1/agentrun_types.go (+13/-4)
- config/crd/bases/sympozium.ai_agentruns.yaml (+12/-27)
- charts/sympozium/crds/sympozium.ai_agentruns.yaml (+12/-27)
- charts/sympozium-crds/templates/sympozium.ai_agentruns.yaml (+12/-27)
- internal/controller/agentrun_controller.go (+11/-0)
- internal/controller/agentrun_sandbox.go (+9/-2)
- test/integration/test-crd-task-polymorphism.sh (+213/-0, new)

## Out of scope (deferred)

The five smaller asks from the earlier review (Prompt() history pollution, structured-output overpromise, prompt_server_test.go tautologies, NewStringTask("") nil semantics, task required-vs-optional) — these are tech debt separate from the schema fix and will be filed for follow-up.

### DIFF
--- a/api/v1alpha1/agentrun_types.go
+++ b/api/v1alpha1/agentrun_types.go
@@ -37,10 +37,19 @@ type AgentRunSpec struct {
 	// by registering a TaskModeHandler — no changes to the central
 	// controller logic required.
 	//
-	// The CRD schema (oneOf: string | object) is hand-maintained in
-	// config/crd/bases/sympozium.ai_agentruns.yaml and the chart template;
-	// kubebuilder cannot express polymorphism directly. The Go side uses
-	// a pointer + custom UnmarshalJSON so the field round-trips cleanly.
+	// The CRD schema is intentionally typeless (no `type:` is emitted)
+	// to accept either a string (legacy Path A) or an object describing
+	// an orchestration mode (Path B). The combination of
+	// `+kubebuilder:validation:Schemaless` (drops the auto-generated
+	// `type: object` plus the unused `properties` block) and
+	// `+kubebuilder:validation:XPreserveUnknownFields` (preserves unknown
+	// fields on the apiserver side) reproduces the polymorphic shape
+	// the field actually has — verified via `--dry-run=server` against
+	// a real apiserver that string-form tasks were previously rejected
+	// by the schema with "spec.task: must be of type object".
+	//
+	// See: PR #302 review (issuecomment 5033007953).
+	// +kubebuilder:validation:Schemaless
 	// +kubebuilder:validation:XPreserveUnknownFields
 	// +kubebuilder:printerColumns:name="Task",type="string",JSONPath=".spec.task"
 	Task *TaskSpec `json:"task,omitempty"`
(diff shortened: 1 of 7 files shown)


AlexsJones (approved):
Thanks for closing this out properly instead of just patching the schema, picking up both of the smaller asks from the #302 review too. `Schemaless` + `XPreserveUnknownFields` is the right combination for a typeless polymorphic field (preserve-unknown-fields alone never relaxed type checking, which is exactly what bit us), all three CRD copies are regenerated consistently, and the nil-task fail-fast plus the sandbox path reaching terminal Failed instead of requeueing forever are both real robustness wins.

One optional follow-up, not blocking: `test/integration/test-crd-task-polymorphism.sh` currently only runs by hand. Wiring it into the Verify workflow behind a kind cluster (or porting it to envtest) would make the admission gap it guards against impossible to reintroduce silently. Happy to take that as a separate PR.

LGTM.

PR #313, opened 2026-07-24
feat: opt out for automatic memories

kristof-ringleff:
Solves the first half of #310 allowing the disabling of the auto memory storing mechanism. Additionally some extra env variables support tweaking the size of the stored memory entries.


AlexsJones (changes requested):
Really solid shape, this is exactly the opt-out #310 asked for: the skill and `/store` stay, only the automatic write stops, and the nil-means-true plumbing keeps every existing agent untouched. The three-level precedence (`agentConfigs[].memory.autoStore` over `autoStoreMemory` over default) with tests at all three layers, the deepcopy fix for `AgentSpec.Memory`, and the docs table are all appreciated.

One change requested before this lands:

**The opt-out never reaches the agentSandbox backend.** `injectMemoryConfig` is called from `reconcilePending` after `buildJob`, but a run with `spec.agentSandbox.enabled` goes through `reconcilePendingAgentSandbox` → `buildSandboxCR` → `buildContainers` and skips that call, so `memory.autoStore: false` silently keeps auto-storing for sandbox-backed runs. Since both backends converge on `buildContainers`, the injection wants to live below that fork (next to where the task-mode hook runs), or be mirrored in the sandbox path. A test asserting the env var lands in the sandbox-built containers too would pin it.

Two small non-blocking notes:

- `injectMemoryConfig` assumes the agent container is `Containers[0]`; if the other inject helpers key off the container name, matching that would be more robust to reordering.
- The doc table says the controller sets `MEMORY_AUTO_STORE` from `memory.autoStore`, worth a half-sentence that setting the env var directly in `spec.agents.default.env` also works and wins today, so users don't file "both set, which one applies" issues later.

With the sandbox path covered this is good to go, and #314 can build on the same seam.

kristof-ringleff:
@AlexsJones thanks for the review. While addressing the shortcomings here in a structural way I stumbled down a rabbit hole. The result is #317 which addresses the divergence issues hopefully on an ongoing basis. Once #317 is merged I'll continue on this PR to bring it back to green.

kristof-ringleff:
rebased on main with #317 now highlighting the parity issue - fixed with [5223ebd](https://github.com/sympozium-ai/sympozium/pull/313/commits/5223ebd16f6d142a3315dd4be215f6edcbc4fa15). [9b8a3c2](https://github.com/sympozium-ai/sympozium/pull/313/commits/9b8a3c2745ef57396749741b59cd4aaa6705b56c) brings the env behaviour inline with other env var behaviour.

AlexsJones (approved):
That was the ask, pulled in cleanly. `injectMemoryConfig` now lives in the pod-mutator list (`internal/controller/agentrun_pod_mutators.go`) instead of the buildContainers fork, so the opt-out reaches both the Job and sandbox paths, and the `hasEnvVar` guard keeps `spec.env` the last word on `MEMORY_AUTO_STORE`. The `memory_auto_store_disabled` parity scenario exercises the sandbox route; `resolveAutoStoreMemory` plus its precedence tests cover the ensemble default and the per-config override.

CI is green and the branch is mergeable now that #317 has landed. Approving for your final look.

kristof-ringleff:
https://github.com/sympozium-ai/sympozium/pull/320 was merged after your review and caused a merge conflict. I have updated this PR in https://github.com/sympozium-ai/sympozium/commit/7691855b2618ba618827f7df8c13e393249e577f to solve the conflict and also use the new agentedit route.

PR #314, opened 2026-07-24
feat: /delete memory (local admin only)

kristof-ringleff:
Provides a /delete endpoint as a back stop for an admin to remove entries. The path is protected by a secret MEMORY_ADMIN_TOKEN_SECRET which is not shared with the agent so only a local admin can perform this action.

See #310

### DIFF
--- a/charts/sympozium/templates/controller-deployment.yaml
+++ b/charts/sympozium/templates/controller-deployment.yaml
@@ -101,6 +101,13 @@ spec:
             - name: AGENT_SANDBOX_DEFAULT_RUNTIME_CLASS
               value: {{ .Values.agentSandbox.defaultRuntimeClass | quote }}
             {{- end }}
+            {{- if .Values.memory.adminDelete.enabled }}
+            # Names the Secret holding the memory-server admin-delete bearer token.
+            # The controller injects it (optional SecretKeyRef) into every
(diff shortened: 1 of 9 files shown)


AlexsJones (approved):
The isolation story here is exactly right: token in a Secret the agents never see, constant-time compare, endpoint hard-disabled when unconfigured, and syncMemoryAdminTokenEnv means enabling it later doesn't require hand-deleting Deployments (on both the per-agent and shared-workflow reconcilers, with the create-only guard pinned by test). The steady-state no-write assertion on resourceVersion is the detail most people forget, and the FTS-trigger check in the delete test proves the row really leaves search, not just the base table.

Two non-blocking notes:

- The `lookup`-based token generation works for imperative helm install/upgrade but renders empty under `helm template` (Argo CD, Flux with post-rendering): every sync would mint a fresh token, and the pod keeps the old value in its env until restart, so your documented rotation caveat triggers continuously. If GitOps installs matter, one doc line steering those users to pre-create the Secret would save a confusing 401 hunt.
- A hard delete can remove a mid-chain entry that other rows reference via parent_id. If getProvenanceChain just stops walking at the gap, that is probably fine for an admin backstop, but worth a sentence or a small test so the behaviour is chosen rather than accidental.

Approving.

PR #315, opened 2026-07-24
feat(apiserver): reload SYMPOZIUM_UI_TOKEN from a mounted file

KSiig:
## Summary

The bearer token was previously bound to the apiserver pod via `valueFrom.secretKeyRef`, which K8s resolves once at pod start. Whenever the Secret rotated the apiserver kept the old value in memory and 401'd every request until a human restarted the pod.

Mount the Secret as a file instead and re-read it on every request so Secret rotations take effect without a pod restart.

## Changes

* New `tokenReader` in `internal/apiserver/token_reader.go` — bearer-token reader that re-reads the file on every `Current()` call. Throttled warn log on read failure; seed path for dev setups that don't deploy the Secret.
* `internal/apiserver/server.go` — `authMiddleware`, `buildMux`, `Start`, `StartWithUI`, `Handler` now take `*tokenReader`. Length-mismatch short-circuit before `subtle.ConstantTimeCompare`.
* `cmd/apiserver/main.go` — added `--token-file` flag (default `/var/run/secrets/sympozium-ui-token/token`); seeds the reader with `--token`/`SYMPOZIUM_UI_TOKEN` as a fallback.
* `charts/sympozium/templates/apiserver-deployment.yaml` — when `apiserver.webUI.token` is empty, replaced the `SYMPOZIUM_UI_TOKEN` env-from-secret with a `volumeMount` + `volume` projecting the `<release>-ui-token` Secret, and added `--token-file=...` as an arg. `defaultMode: 0444` so the nonroot apiserver (UID 65532 in distroless) can read the file.
* 16 `test/integration/test-api-*.sh` + `test-memory-persistence.sh` — `resolve_apiserver_token` extracted to `test/integration/lib/resolve-token.sh`; each script `source`s the helper. `test-apiserver-token-reload.sh` does the same.
* `test/integration/test-apiserver-token-reload.sh` (new) — full live-cluster test: initial auth, two consecutive Secret rotations, pod `StartTime` check to assert no pod restart. `wait_for_token` polls the apiserver with the new bearer token and breaks on the first non-401 response.
* New unit tests in `internal/apiserver/token_reader_test.go` and `internal/apiserver/auth_test.go` cover the reader and the middleware contract. `TestTokenReader_RotationWithSameMtime` is a regression test for the kubelet mtime-stability issue.

## Verification

* `go build ./...` — clean
* `go vet ./...` — clean
* `gofmt -l .` — clean
* `go test -count=1 -timeout 300s ./...` — every package passes
* `bash test/integration/test-apiserver-token-reload.sh` against a live cluster:
  * Initial request: HTTP 200
  * Round 1: new token HTTP 200, old token HTTP 401
  * Round 2: new token HTTP 200, old token HTTP 401
  * Apiserver pod `StartTime` unchanged across rotations

## Deviations from the original spec

1. **`defaultMode: 0444`, not `0400`.** The distroless apiserver runs as UID 65532 (nonroot); K8s Secret volume mounts are owned by root, so `0400` makes the file unreadable by the apiserver and 401s every request. `0444` is the smallest readable mode that works with `runAsNonRoot: true`.
2. **Per-request file re-read, not an mtime cache.** A previous implementation used an mtime fast-path; it was removed because kubelet does not always bump the projected file's mtime on rapid Secret patches (observed as 60-90s of drift on the dev cluster). An mtime cache could serve a stale token for that window. Cost is one stat + read per request (a few µs). The `TestTokenReader_RotationWithSameMtime` regression test exercises the exact scenario.
3. **Test wait budget 120s** for kubelet projection. The test polls the apiserver with the new bearer token inside the budget and breaks on success.

## Operational notes

* **One-time post-upgrade step (reflector):** HelmRelease post-renderers that add annotations to the `sympozium-ui-token` Secret (e.g. Reflector's `reflection-allowed`) need to re-run after the chart upgrade. If they don't, `kubectl rollout restart deploy/sympozium-apiserver` picks the annotation up on the new pod. The `apiserver-ui-secret.yaml` template's `helm.sh/resource-policy: keep` is unchanged, so the Secret itself is preserved across upgrades.
* The `sympozium.fullname` helper continues to derive the Secret name, so any chart `nameOverride` keeps working.

## Out of scope

* Hot-reload for other apiserver-affecting secrets (model-provider credentials, etc.) — different lifecycle, not affected by this bug.
* Switching from a 32-byte `randAlphaNum` token to a longer or signed token — separate from "make the existing token live-reload".
* Replacing the Secret-based token with a JWT or mTLS — the right long-term move but separate from this fix.

### DIFF
--- a/charts/sympozium/templates/apiserver-deployment.yaml
+++ b/charts/sympozium/templates/apiserver-deployment.yaml
@@ -27,6 +27,7 @@ spec:
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           args:
             - --event-bus-url={{ include "sympozium.natsUrl" . }}
+            - --token-file=/var/run/secrets/sympozium-ui-token/token
           {{- if .Values.apiserver.webUI.enabled }}
             - --serve-ui
           {{- end }}
@@ -34,13 +35,6 @@ spec:
           {{- if .Values.apiserver.webUI.token }}
             - name: SYMPOZIUM_UI_TOKEN
               value: {{ .Values.apiserver.webUI.token | quote }}
-          {{- else }}
-            - name: SYMPOZIUM_UI_TOKEN
-              valueFrom:
-                secretKeyRef:
-                  name: {{ include "sympozium.fullname" . }}-ui-token
-                  key: token
-                  optional: true
           {{- end }}
             - name: POD_NAME
               valueFrom:
@@ -100,6 +94,12 @@ spec:
             {{- toYaml .Values.apiserver.resources | nindent 12 }}
           securityContext:
             {{- toYaml .Values.containerSecurityContext | nindent 12 }}
+          {{- if not .Values.apiserver.webUI.token }}
+          volumeMounts:
+            - name: sympozium-ui-token
+              mountPath: /var/run/secrets/sympozium-ui-token
+              readOnly: true
+          {{- end }}
       {{- with .Values.apiserver.nodeSelector }}
       nodeSelector:
         {{- toYaml . | nindent 8 }}
@@ -115,6 +115,17 @@ spec:
       {{- with .Values.imagePullSecrets }}
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- if not .Values.apiserver.webUI.token }}
+      volumes:
+        - name: sympozium-ui-token
+          secret:
+            secretName: {{ include "sympozium.fullname" . }}-ui-token
+            # 0444 (readable by all) is required because the apiserver runs
+            # as non-root (distroless nonroot: UID 65532) and K8s Secret
+            # volume mounts are owned by root. 0400 would only be readable
+            # by root and the apiserver would 401 every request.
+            defaultMode: 0444
       {{- end }}
 ---
 apiVersion: v1
--- a/charts/sympozium/templates/apiserver-deployment.yaml
+++ b/charts/sympozium/templates/apiserver-deployment.yaml
@@ -94,7 +94,8 @@ spec:
             {{- toYaml .Values.apiserver.resources | nindent 12 }}
           securityContext:
             {{- toYaml .Values.containerSecurityContext | nindent 12 }}
-          {{- if not .Values.apiserver.webUI.token }}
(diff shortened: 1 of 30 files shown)


AlexsJones (changes requested):
This closes a real operational trap (rotate the Secret, everything 401s until someone bounces the pod), and the test coverage is the kind I want: rotation, permission-denied fallback with the cached value, the ws query path, and the length short-circuit all pinned.

Two things to fix before merge:

1. **The middleware fails open if the token file is rotated to an empty value.** `buildMux` decides auth-on once at startup, but `Current()` is re-read per request: write an empty token into the Secret and a request with no Authorization header at all now passes, because `len("") == len("")` survives the length check and `subtle.ConstantTimeCompare` on two empty slices returns 1. A failed read keeps the last known value, so it takes a deliberate (or buggy) empty write to trigger, but auth should fail closed: 401 up front in the middleware when `expected.Current()` is empty. Running open is a startup decision, not something a runtime rotation should be able to flip.

2. **The new Secret volume can strand the pod.** `apiserver-ui-secret.yaml` only renders when `webUI.enabled` is true and no literal token is set, but the volume you add to the deployment is gated only on the literal token being unset. A headless install (`webUI.enabled: false`, no token) mounts a Secret that never exists and the pod sits in ContainerCreating. The env path you removed had `optional: true` for exactly this case. Either add `optional: true` to the secret volume (the reader already tolerates a missing file, and `Seed` covers the fallback) or gate the volume identically to the Secret template.

Nits, non-blocking: `authMiddleware`'s comment says `Current()` is "a single stat() when the mtime has not changed", but the reader deliberately always re-reads (its own comment cites the kubelet mtime issue), and the struct's `mtime` field is written but never read; `Seed`'s comment still references the `optional: true` chart semantics this PR removes. Worth a tidy pass so the next reader isn't misled.

AlexsJones (approved):
Both asks landed clean. The middleware now fails closed, an empty expected token rejects any bearer (401 with explicit tests for no-header, empty-bearer, and any-bearer), and the helm template gates the Secret volume mount and volumes on the token existing, so headless installs without a Secret aren't stranded behind a non-optional mount. Verify green and the branch is mergeable. Good to go.

PR #316, opened 2026-07-29
feat: compact old tool results

kristof-ringleff:
I believe there are some token efficiencies to be gained by compacting tool results from previous rounds. The savings have to be bigger though than the invalidated previously cached prefixes this produces - as such I have included an initial plumbing of keeping track of cached reads (openai is functional for now as the other providers I think do not yet opt into it).

The compacting behaviour is opted in and manageable via environment variables:

Variable | Default | Effect
-- | -- | --
CONTEXT_TOOL_RESULT_MAX_BYTES | 0 (off) | Truncates each tool result as it enters the conversation.
CONTEXT_HISTORY_BUDGET_BYTES | 0 (off) | Once live tool-result bytes exceed this, older results are replaced with short placeholders.
CONTEXT_HISTORY_BUDGET_LOW_BYTES | half the budget | How far down to drain when elision fires.
CONTEXT_KEEP_RECENT_RESULTS | 3 | Newest results never elided — the model is still reasoning over these. The in-flight round's results are protected on top of this window. Only applies when elision is on.

I am still experimenting with these values and I think there will be an ongoing need to tweak this based on context size and shape of work (ie a single fetch_url can already be 50K) without a one size fits all approach embedded in the defaults. Future work could apply some smarts against the context size.

To be applied after #312

### DIFF
--- a/cmd/agent-runner/context_policy.go
+++ b/cmd/agent-runner/context_policy.go
@@ -120,13 +120,22 @@ func jsonBytes(v any) int {
 //
 // The cut backs off to a rune boundary: a byte-exact slice can split a
 // multi-byte rune, and json.Marshal would then silently substitute U+FFFD into
-// the request. Same guard the MCP output cap uses (mcp_tools.go).
+// the request. Only the trailing rune is examined: validating the whole kept
+// prefix would be quadratic, and a single invalid byte anywhere earlier — the
+// first byte of binary execute_command output, say — would rewind the cut back
+// to it and throw away everything after.
 func (cp contextPolicy) clampToolResult(tool, content string) (string, int) {
 	if cp.ToolResultMaxBytes <= 0 || len(content) <= cp.ToolResultMaxBytes {
 		return content, 0
 	}
 	kept := content[:cp.ToolResultMaxBytes]
-	for len(kept) > 0 && !utf8.ValidString(kept) {
+	// A rune split by the cut orphans at most UTFMax-1 bytes, so the back-off
+	// is bounded. Content that is invalid at the cut for any other reason is
+	// left alone rather than chased backwards.
+	for i := 0; i < utf8.UTFMax-1 && len(kept) > 0; i++ {
+		if r, size := utf8.DecodeLastRuneInString(kept); r != utf8.RuneError || size > 1 {
+			break
+		}
 		kept = kept[:len(kept)-1]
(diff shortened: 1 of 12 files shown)


kristof-ringleff:
Example measurement from a long running agent process (200+ rounds)


1,781,198 normal-token-equivalents — 46.1% cheaper.

|                 | cost (in uncached-input-token equivalents) |
|-----------------|--------------------------------------------|
| without elision | 3,861,094                                  |
| with elision    | 2,079,896                                  |
| **saved**       | **1,781,198**                              |

AlexsJones (changes requested):
This is careful work, and the invariant handling is exactly what I'd want: results are swapped in place keyed by call id on all three providers and never removed, so an orphaned tool_result can't happen, and the off-by-default posture plus the current-round protection are both pinned by tests. The batched drain to a low-water mark is the right shape for cache behaviour too.

One structural blocker: the diff carries all of #312 as byte-identical hunks across seven files (CRD task fix, admission check, sandbox failRun). #312 is approved and in my merge queue, so let's land that first and rebase this down to `cmd/agent-runner/*` plus docs. Merging as-is would silently fold #312 in, and any late change there forces a rebase here anyway.

Two small ones for the rebase:

- `clampToolResult`'s rune backoff runs `utf8.ValidString` over the whole kept string, so invalid UTF-8 anywhere before the cap (think binary `execute_command` output) trims everything back to the first invalid byte, and it's quadratic besides. Bound the backoff to the trailing rune (`utf8.DecodeLastRuneInString`) instead.
- docs/guides/context-cost.md says native and MCP tool output "are already capped at 8 KB each" but earlier notes `fetch_url` caps at 50 KB; one of those needs correcting.

And a one-liner worth adding: a comment in the Anthropic `ReplaceToolResults` noting the rebuild is deliberately text-only, so it gets caught if tool_result content ever grows image blocks.

kristof-ringleff:
rebased with review notes addressed

AlexsJones (approved):
The rebase cleared it: this is down to cmd/agent-runner/* plus docs now, no more carried #312 content, and the rune-backoff fix is in (utf8.DecodeLastRuneInString with tests for whole-rune cuts, invalid-UTF-8-before-the-cap, and the zero-cap disable). Verify green and mergeable. Good to go.

PR #317, opened 2026-07-30
 refactor(controller): converge the AgentRun backend and Ensemble create/update paths + parity tests

kristof-ringleff:
## Summary

Two places in the controller fork into parallel paths that are meant to produce the same
result, and each had drifted because every new field had to be remembered twice:

- **AgentRun → pod**: `reconcilePending` builds a `batchv1.Job`; `reconcilePendingAgentSandbox`
  builds an agent-sandbox `Sandbox` CR. Each shaped its own pod spec.
- **Ensemble → Agent**: `reconcileAgentConfig` creates via `buildAgent`, or updates an
  existing Agent field by field.

Both now converge on a single builder, and tests hold them there.
`internal/controller/agentrun_sandbox.go` previously had no test coverage.

## Prior occurrences

| | |
|---|---|
| `46250af` (#67) | Lifecycle hooks: `ensureLifecycleRBAC`, workspace PVC, and `startPostRun` transitions added to `reconcilePending`, then separately to `reconcilePendingAgentSandbox` |
| `99ddb4d` | Making skill/lifecycle RBAC failure fatal had to be applied in both pending paths |
| `e33be9f` (#264) | Ensemble create/update: `authRefs`, `model`, `volumes`/`volumeMounts`, `agentSandbox`, `lifecycle` set on create, not propagated on update |
| #313 (review) | `injectMemoryConfig` called from `reconcilePending` only, so `memory.autoStore: false` did not reach sandbox-backed runs. Not on `main` yet; once this lands it registers as a `podMutator` and is covered |

Each was fixed by mirroring logic into the second path. This removes the second path.

## Changes

**Shared pod template.** `buildAgentPodTemplate` owns everything below the backend fork —
`buildContainers`, `buildVolumes`, host access, pod security, labels — and applies the
`podMutators` registry. `buildJob` wraps it in a `JobSpec`; `buildSandboxCR` converts the same
template. Shared reconcile prerequisites move into `resolveAgentRunInputs` /
`resolveProviderHeaders`. The three `inject*` methods move to `agentrun_pod_mutators.go`, take
`*corev1.PodSpec`, and locate the agent container by name rather than `Containers[0]`.

**Whole-spec Agent updates.** The update branch of `reconcileAgentConfig` now builds desired via
`buildAgent` and assigns the whole spec, matching the pattern the schedule fork in the same file
already used. Labels keep merge semantics. Net −245 lines; three comparators that existed only to
diff individual fields are gone, and `buildAgent` now calls `buildDesiredSkills` instead of
duplicating it.

**Converters.** `containerToMap` / `podSpecToMap` stay explicit allowlists, guarded by reflection
tests that fail on any field neither converted nor listed in an omit table with a reason.
`volumeToMap` is now lossless: `buildVolumes` appends `agentRun.Spec.Volumes` verbatim, so its
input is user-authored and the `VolumeSource` union spans the core API — an unrecognised source
produced a volume with no source, which the apiserver rejects.

**Shared prerequisites.** The ~150 lines of setup either path needs before rendering a pod —
ServiceAccount, input and MCP ConfigMaps, traceparent, sidecar resolution, sidecar-tools manifest,
memory-readiness gate, RBAC, workspace PVC, provider headers — were duplicated and are now
`prepareRunPrerequisites` + `prepareTaskPrerequisites`. The split falls where the Job path branches
to server mode: the sidecar-tools ConfigMap is only mounted by task-mode pods, so writing it before
that branch would orphan one on every server-mode run. `TestServerModeLeavesNoSidecarToolsConfigMap`
pins that ordering. Each reconcile function is now its own backend's build-and-create step plus two
shared calls.

That closes the one residual divergence: the Job path built the sidecar-tools manifest from the
unfiltered sidecar list while the sandbox path used the `RequiresServer`-filtered one, so a
server-only sidecar declaring tools advertised them to a task-mode agent whose sidecar was absent
from the pod.

`resolveAgentRunInputs` returns an error for non-`NotFound` Get failures rather than rendering a pod
with memory, observability, skills, ensemble label and timeout silently dropped.
`buildSandboxClaimCR` shares `agentPodLabels` instead of hand-writing the set.

## What this closes

Sandbox-backed runs:

- `env[].valueFrom` was not converted, so the provider API key (`SecretKeyRef` over
  `allowedAuthSecretKeys`), canary `HOST_IP`, and MCP auth env did not reach the pod. The sandbox
  integration tests use a local no-auth model.
- `seccompProfileForPod` was not called, so pods carried no seccomp profile. Same in
  `buildPostRunJob`.
- Shared memory, relationship context, and subagents config were not applied; the
  `sympozium.ai/ensemble` label was not propagated either, which gates two of them.
- The sidecar-tools ConfigMap was mounted by `buildVolumes` but created only by the Job path.
- `spec.timeout`, `providerHeadersSecretRef`, and the observability `ResourceAttributes` deep copy
  were absent; the last aliased the Agent's map.
- Container `securityContext` fields used by host-access sidecars (`privileged`,
  `seccompProfile`, `runAsUser`) were dropped, as were user volumes with a source outside
  `emptyDir`/`configMap`/`projected`/`persistentVolumeClaim`.

Ensemble-managed Agents: `displayName`, `observability`, `memory.enabled`/`maxSizeKB`, and the
`sympozium.ai/ensemble` / `sympozium.ai/agent-config` labels were not propagated on update. The
labels matter beyond cosmetics — `toolpolicy.ForAgent` returns `nil` without them (dropping the
persona's `toolPolicy`), `injectSharedMemory` falls back to `read-write`, and
`injectRelationshipContext` returns early. All three fail open.

## Tests

Both suites drive the real reconcile entry points and compare what was persisted, since the
divergence sits at the call site rather than in a builder. Shared helpers live in
`parity_helpers_test.go` (`diffStructs`, `fillStruct`).

- `agentrun_backend_parity_test.go` — 11 scenarios must render identical pod specs, compared field
  by field with collections keyed by name. Each scenario also asserts its feature is present on
  both backends independently, so a fixture that stops triggering its feature fails rather than
  comparing absent to absent. `parityBaseline` is empty. Tier 2 covers the paths that differ by
  design (postRun Job, sandbox `runtimeClassName`, SandboxClaim).
- `ensemble_parity_test.go` — convergence against `buildAgent`'s output as the oracle: a perturbed
  Agent must converge, an edited Ensemble must reach existing Agents, and every `AgentSpec` field
  is either derived from the Ensemble or declared inexpressible with a reason.
- `agentrun_sandbox_convert_test.go` — converter completeness and omit-table hygiene.
- `TestBackendParity_SidecarToolsConfigMap` — side-resource parity, since a divergence in ConfigMap
  *contents* is invisible to a pod-spec diff.
- `TestAgentSpecHasNoServerSideDefaults` (envtest) — the whole-spec comparison converges only while
  the apiserver returns a spec byte-identical to what was submitted, so this fails if a
  `+kubebuilder:default` lands on an `AgentSpec` field `buildAgent` leaves zero. The fake client
  applies no defaults, so no unit test can cover it.

- `TestServerModeLeavesNoSidecarToolsConfigMap` — the ordering constraint behind the two-helper
  split, since folding them into one is the obvious simplification and silently orphans a ConfigMap.

Thirteen negative controls were run to confirm the guards fire, including reproducing #313's exact
shape (mutators wired into the Job path only), #264's (update reverted to field-by-field), the
unfiltered sidecar list, a real `+kubebuilder:default` added to `AgentConfig.Thinking`, and moving
the server-mode branch past the task-mode setup.

## Notes

- **Behaviour change:** `thinking`, `sandbox`, `nodeSelector`, `webEndpoint`, and
  `imagePullSecrets` are cleared on the next reconcile of an Ensemble-managed Agent, since
  `buildAgent` never sets them and `AgentConfigSpec` cannot express them. They are declared in
  `agentFieldsNotExpressibleByEnsemble` with reasons. Making one configurable means adding it to
  `AgentConfigSpec`, not adding an exception.

  **This has a concrete victim.** The TUI's agent-edit flow (`cmd/sympozium/main.go:5275`) writes
  `spec.webEndpoint` and `spec.memory.enabled` directly onto existing Agents. On an
  Ensemble-managed Agent that used to stick and now reverts. The controller logs
  `"Reverting out-of-band change to ensemble-managed Agent"` with the instance, ensemble, and a
  hint to edit `spec.agentConfigs` instead, so this is diagnosable rather than a silent
  "the toggle doesn't work". A TUI guard on Agents carrying the `sympozium.ai/ensemble` label is
  the natural follow-up.
- `spec.agentSandbox.warmPoolRef` appears non-functional independently of this change:
  `buildSandboxClaimCR` emits only `warmPoolRef` with no pod spec, and
  `reconcileRunningAgentSandbox` then looks up a `Sandbox` by the claim's name.
  `TestBackendInvariants/sandboxClaimCarriesNoPodSpec` pins current behaviour; worth a separate
  issue.
- `go build`, `go vet`, `gofmt`, and `make test` (race) pass. `make test-system` fails
  `TestAgentCreateViaAPI`, `TestRunCreateDelete`, and `TestRunJobShape`, all of which fail
  identically on `main`. Two are a genuine CRD schema mismatch (`spec.task in body must be of type
  object`). `TestAgentCreateViaAPI` is a different, pre-existing bug: `assertExists`
  (`test/system/helpers_test.go:85`) does a single un-polled Get against `mgr.GetClient()`, the
  *cached* client, immediately after a create — an informer race. The passing tests all use
  `pollUntil`. Worth a separate fix.

### DIFF
--- a/internal/controller/agentrun_backend_parity_test.go
+++ b/internal/controller/agentrun_backend_parity_test.go
@@ -1,0 +1,888 @@
+package controller
+
+import (
+	"context"
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"os"
+	"sort"
+	"strings"
+	"testing"
+
+	"github.com/go-logr/logr"
+	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	sympoziumv1alpha1 "github.com/sympozium-ai/sympozium/api/v1alpha1"
+)
+
+// ── backend parity ────────────────────────────────────────────────────────────
+//
+// An AgentRun runs as either a batchv1.Job (reconcilePending) or an agent-sandbox
+// Sandbox CR (reconcilePendingAgentSandbox). Both are expected to produce the same
+// pod.
+//
+// These tests enter at the reconcile functions rather than the builders. The
+// builders are shared, so divergence arises at the call site — in which
+// prerequisites and mutators each path applies. The assertions compare what each
+// backend persisted: a Job in the fake client, a Sandbox CR in the fake dynamic
+// client.
+//
+// Tier 1 (TestBackendParity) requires identical pod specs from the two task-mode
+// backends. Tier 2 (TestBackendInvariants) covers the paths that differ by design
+// — postRun Job, sandbox runtimeClassName, SandboxClaim — with per-path
+// invariants.
+
+// parityBaseline suppresses accepted divergences between the Job and Sandbox
+// backends.
+//
+// Key: dotted field path as produced by diffStructs, optionally ending in "*".
+// Value: reason, prefixed "intentional:" or "known-gap:" with an issue number for
+// the latter.
+//
+// Currently empty: both backends render from buildAgentPodTemplate. An added entry
+// means the two ship different pods and needs sign-off in review.
+// TestParityBaselineHasNoStaleEntries fails entries that no longer match a
+// divergence.
+var parityBaseline = map[string]string{}
+
+// ── Tier 1: strict Job ↔ Sandbox parity ───────────────────────────────────────
+
+func TestBackendParity(t *testing.T) {
+	for _, sc := range parityScenarios() {
+		t.Run(sc.name, func(t *testing.T) {
+			if sc.guards != "" {
+				t.Logf("guards: %s", sc.guards)
+			}
+
+			jobSpec := jobBackendPodSpec(t, sc)
+			sandboxSpec := sandboxBackendPodSpec(t, sc)
+
+			// A fixture that stops triggering its feature renders it as absent on
+			// both backends, leaving the diff below empty. Assert presence on each
+			// side first.
+			assertScenarioIsLive(t, "job backend", sc, jobSpec)
+			assertScenarioIsLive(t, "sandbox backend", sc, sandboxSpec)
+
+			diffs := diffStructs(t, "spec", &jobSpec, &sandboxSpec)
+
+			for _, d := range diffs {
+				if reason, ok := matchBaseline(d.path); ok {
+					t.Logf("known divergence at %s: %s", d.path, reason)
+					continue
+				}
+				t.Errorf("backend divergence at %s\n  job backend:     %s\n  sandbox backend: %s\n\n"+
+					"Both backends render from buildAgentPodTemplate; apply shared pod changes there "+
+					"or register a podMutator.",
+					d.path, d.a, d.b)
+			}
+		})
+	}
+}
+
+// assertScenarioIsLive checks the scenario's feature landed on the agent
+// container, so parity is compared over a pod that carries it.
+func assertScenarioIsLive(t *testing.T, side string, sc parityScenario, spec corev1.PodSpec) {
+	t.Helper()
+
+	agent := agentContainer(&spec)
+	if agent == nil {
+		t.Fatalf("%s: scenario %q rendered no %q container", side, sc.name, agentContainerName)
+	}
+
+	present := make(map[string]corev1.EnvVar, len(agent.Env))
+	for _, e := range agent.Env {
+		present[e.Name] = e
+	}
+	for _, want := range sc.wantAgentEnv {
+		e, ok := present[want]
+		if !ok {
+			t.Fatalf("%s: scenario %q expects env %s on the agent container; it is absent.\n"+
+				"Either the fixture no longer triggers the feature, or the feature regressed on "+
+				"this backend.", side, sc.name, want)
+		}
+		if e.Value == "" && e.ValueFrom == nil {
+			t.Errorf("%s: scenario %q env %s is present but carries neither value nor valueFrom",
+				side, sc.name, want)
+		}
+	}
+	for _, want := range sc.wantInitContainers {
+		if !hasContainerNamed(spec.InitContainers, want) {
+			t.Fatalf("%s: scenario %q expects init container %q; got %v",
+				side, sc.name, want, containerNames(spec.InitContainers))
+		}
+	}
+	for _, want := range sc.wantVolumes {
+		if !hasVolumeNamed(spec.Volumes, want) {
+			t.Fatalf("%s: scenario %q expects volume %q; got %v",
+				side, sc.name, want, volumeNames(spec.Volumes))
+		}
+	}
+}
+
+func hasContainerNamed(list []corev1.Container, name string) bool {
+	for _, c := range list {
+		if c.Name == name {
+			return true
+		}
+	}
+	return false
+}
+
+func hasVolumeNamed(list []corev1.Volume, name string) bool {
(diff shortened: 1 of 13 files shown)


AlexsJones (commented):
This is the rare refactor where the tests are the headline: reflection guards over every corev1 field, the AST scan for unregistered mutators, and parity scenarios that enter at `reconcilePending` mean the next divergence fails a test instead of shipping. The bug closures along the way are real too (sandbox pods regaining `valueFrom` env, seccomp, and container securityContext, the phantom sidecar-tools ConfigMap, server mode no longer orphaning one).

The one thing I want to sit with before merging is the Ensemble update branch. Whole-spec assign makes the Ensemble authoritative, so out-of-band edits to managed Agents (including the TUI edit flow you flagged) revert on the next reconcile. You've documented it and the log line makes it diagnosable, and converge-to-declared is what a controller should mean, but it changes effective behaviour for anyone relying on TUI-set `webEndpoint` or `memory.enabled`, so I want to make that call explicitly rather than by merge. Non-blocking notes either way:

- `resolveAgentRunInputs` erroring on non-NotFound Get failures instead of rendering a degraded pod is right, but it is a reconcile behaviour change, worth a release-note line.
- Tier-1 parity scenarios don't yet cover lifecycle preRun hook containers or providerHeaders env; one fixture each would give the reflection guards something to bite on there.
- Several unit tests drive mutators with a nil-client reconciler; safe today because every mutator gates before `r.Get`, but a future mutator that reads first will panic in tests. A one-line doc comment on the registry would spare someone the surprise.

Review on #316 is up as well; merge order as I see it is #312, then this, then #316 rebases down to its own files.

kristof-ringleff:
@AlexsJones Thanks for the review. Confirming on the merge order #312 is needed as currently anything branching off main leads to non-working tasks making local testing tricky when merge conflicts come up.

The ability to edit ensemble controlled agents directly currently opens up another avenue of divergence causing confusion. Editing an Agent directly then going back to the Ensemble config not seeing those changes.

The fix for both issues (removing functionality and diverging edits) is to land the edit on the ensemble controlled agent directly on the ensemble. However as of now this is not possible as not all values expressible on the agent are also available on the ensemble (agent to agentconfig parity). 

I have added a commit that addresses the review notes first and then separate commits that keep the previous functionality with the agent edit to ensemble redirect. Happy to split this off into a separate follow up PR if you prefer.

AlexsJones:
Ruling on the open question: Ensemble-authoritative is what I want. Converge-to-declared is the contract a controller should keep, the revert is logged and diagnosable, and a carve-out list of "TUI-editable" fields would rot within a release or two. So whole-spec assign stays.

Your read on the real fix matches mine: if editing a managed Agent matters, the edit belongs on the Ensemble. The redirect commits (91c8dfc, 775f908) landed after my review so I haven't been through them yet, I'll do that on the next pass. If they turn out to be more than a thin redirect, splitting them into a follow-up is fine by me and keeps this PR about the convergence refactor. Agent to AgentConfig parity is worth its own issue either way.

One thing before this can land: the branch has drifted to conflicting with `main`. #312 is merged now, so a rebase on top of that should be the clean way through. Push it and I'll take the final look. #313 and #316 come right behind.

kristof-ringleff:
@AlexsJones sounds good - I moved the new commits handling the agent edit redirect to PR #320 and solved the merge conflict here.

AlexsJones (approved):
Rebase is clean now: merged upstream main (incl. #312), branch is MERGEABLE, redirect work relocated to its own follow-up #320 so this PR is the convergence alone. Verify and the parity/sandbox/roundtrip tests are green. This is the block that unblocks #313 and #316, so approving it for your final look rather than sitting on the chain.

PR #320, opened 2026-08-05
feature: agent edit redirect

kristof-ringleff:
Follow up to #317 

Any ensemble controlled agent edits should land on the ensemble's agent config. As not all fields were previously expressible via the ensemble the missing fields are added to the ensemble's CRD first.


AlexsJones (approved):
This is the redirect done right. Routing each edit to the owning Ensemble's agentConfigs entry (with the orphan-label fallback to standalone) is the honest implementation of the ensemble-authoritative ruling from #317, and packing the routing rule in one place in `internal/agentedit` keeps the TUI and the API on the same path.

The parts that stood out: `reflect.DeepEqual` skips a pointless write so a re-saved-unchanged form neither hits the API nor blocks the reconcile wait, and the `Observed`/reconcileWait handshake means the TUI no longer seeds its edit form from a read that can race the controller. The inverse mapping between `applyToAgentConfig` and `buildAgent`/`buildDesiredSkills` is the kind of coupling that rots silently, so pinning it with the round-trip test and the web-endpoint derived-params guard is exactly right. `Apply` returning `Target` and both callers surfacing "Applied via …" means the user is told where the change landed.

CI is green and the branch is mergeable. Approving for your final look; this closes out the last behavioral edge from the #317 decision.

PR #337, opened 2026-08-17
Make elided output directly recoverable via read_file

kristof-ringleff:
Tool results can potentially take a while to determine (think test run output) or can be time bound. Rather than completely wiping the content we can store it on disk and provide instructions to read it back if needed rather than having to re-run the tool. This provides an approach to have Context addressable via a variable.

### DIFF
--- a/cmd/agent-runner/context_policy.go
+++ b/cmd/agent-runner/context_policy.go
@@ -5,7 +5,9 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"path/filepath"
 	"strconv"
+	"strings"
 	"unicode/utf8"
 )
 
@@ -17,6 +19,12 @@ const (
 	// as it did before unless one of the CONTEXT_* vars is set.
 	defaultToolResultMaxBytes = 0
 	defaultKeepRecentResults  = 3
+
+	// defaultElisionSpillDir is where elided tool output is parked so the model
+	// can read it back. It sits under the same /workspace/.sympozium prefix the
+	// runner already uses for trace-context.json, and is only ever created when
+	// elision actually fires.
+	defaultElisionSpillDir = "/workspace/.sympozium/elided"
 )
 
 // contextPolicy bounds how much tool-result content accumulates in the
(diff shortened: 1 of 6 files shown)


kristof-ringleff:
ping @AlexsJones when you get a chance can you please assess this PR alongside the others - cheers

AlexsJones (changes requested):
Thanks for the ping, and sorry this sat. Spilling elided output to disk is a good idea, and the details are well thought through: the stub points at a real file, spilling is skipped when `read_file` isn't available or can't reach the directory, and the filename comes only from the ledger index. Two things before it can go in:

- The branch is 523 commits behind main and its only `Verify` run is from 17 Aug. `runAgentLoop` gains a parameter here, so please rebase so CI checks every current caller.
- The default spill dir is `/workspace/.sympozium/elided`, and `/workspace` is the persistent volume for workspace sessions. Elided output (which can hold anything a command echoed) would outlive the run there, and the next run's `0000.txt` overwrites the previous run's. Scoping it per run (`AGENT_RUN_ID` is already in the env) or defaulting to `/tmp`, which is also a readable root, would avoid both.

Small nit: `mockTools` in `provider_test.go` now sits between `mockProvider`'s doc comment and the type, so that comment documents the var instead.

kristof-ringleff:
@AlexsJones Thanks for the review comments - these have now been addressed with tmp being the default. I have also rebased on main for the latest changes.

AlexsJones (approved):
Thanks for the rebase and the quick follow-up! Both points are covered: the spill dir now defaults to `/tmp/sympozium/elided`, and `runScopedSpillDir` adds the `AGENT_RUN_ID` subdirectory, with a check that a crafted ID can't climb out of the base. `/tmp` is a readable root in `paths.go` and is mounted as an emptyDir in the agent container, so spills are readable and go away with the pod. All four `runAgentLoop` callers pass the tool list after `applyToolPolicy`, so the `read_file` availability check sees the real tool set. Verify is green on d838b5c, and the `mockTools` comment placement is fixed too. Good to merge.

PR #338, opened 2026-08-17
feat: Add gate-driven retry to response gates

kristof-ringleff:
A response gate can now return `{"action":"retry","reason":…,"response":…}` and get another attempt instead of ending the run: the controller creates a successor AgentRun whose task is a ## Retry N of M card carrying the original task, the rejected output, and the gate's feedback, while the superseded attempt ends Failed with gateVerdict: retried and publishes nothing.


AlexsJones (changes requested):
Thanks for your patience on this one, and for documenting the retry semantics so carefully in `lifecycle-hooks.md`. Two things before a full review makes sense:

- The branch is 525 commits behind main and its only Verify run is from August. `agentrun_controller.go` has moved a lot since then (gated completion, channel failure replies, new phases), so please rebase so CI checks the retry path against current code.
- The postRun timeout fix stands on its own. Main still measures the postRun budget from `Status.StartedAt` in `reconcilePostRunning`, so any gated run longer than ten minutes has its gate killed on the first PostRunning reconcile. Could you split `postRunJobStart`/`postRunBudget` and `agentrun_postrun_timeout_test.go` into their own small PR? That can land quickly while the retry design (the new `retry` gate action, successor AgentRuns, and the CRD and policy fields) gets a proper look.

#339 builds on this, so I'll pick it up once this one is rebased.

kristof-ringleff:
@AlexsJones this is now rebased with the timeout changes moved to a separate PR

AlexsJones (changes requested):
Thanks for rebasing and moving the timeout fix to #589, the retry flow is much easier to review on its own. The deterministic successor names, the lineage labels as a fallback for a lost status write, and resolving an un-retryable verdict as reject rather than approve all look right. Two things before this can land:

- **Delegated runs.** If a delegate child's gate returns `retry`, `retireForRetry` marks the child `Failed` with an empty `status.error`. The parent's `reconcileAwaitingDelegate` still points at the old `ChildRunName`, so it records that delegation as failed while the successor is still running. The successor doesn't carry the parent linkage either, so its result never reaches the parent. Either repoint the parent's delegate entry at the successor and copy the linkage, or have admission reject `lifecycle.retry` on delegated runs for now. A test with a delegated child that retries once and then passes would pin it.
- **Timeout docs.** The new `LifecycleHookContainer.Timeout` comment (and the regenerated CRD descriptions) describe summed postRun budgets, which only exist once #589 lands. Could you move that doc change into #589, or say here that this PR depends on it?

One question too: retired attempts sit at phase `Failed`. Does anything else react to `Failed` (sequential successors, schedule status, run-history pruning) that should skip `gateVerdict: retried`?

kristof-ringleff:
> One question too: retired attempts sit at phase `Failed`. Does anything else react to `Failed` (sequential successors, schedule status, run-history pruning) that should skip `gateVerdict: retried`?

Good question — it turned up more than the delegation case. Here is every site that reads `phase: Failed`, with where the branch now stands. Rows 3–10 were broken and are fixed here; 15–18 are deliberate; 19–23 are listed so the answer is complete rather than filtered.

| # | Site | Reads a retired attempt as… | Status |
|---|---|---|---|
| 1 | `failRun` → `TopicAgentRunFailed` (`agentrun_controller.go:5113`) | nothing — `retireForRetry` never publishes | ✅ by design |
| 2 | Channel router (`channel_router.go:50`) | nothing — subscribes to *completed* only; the successor carries the reply address | ✅ by design |
| 3 | `reconcileAwaitingDelegate` (`agentrun_controller.go:1472`) | ~~a failed delegation → fails the parent~~ | ✅ **fixed** — `reassignDelegation` (`agentrun_retry.go:254`) repoints `status.delegates[].childRunName` before the retire |
| 4 | SpawnRouter result delivery (`spawn_router.go:82` `claimPending`) | ~~an unknown run → result never reaches the parent~~ | ✅ **fixed** — walks the lineage back to the registered key |
| 5 | Subagent batch slots (`spawn_router.go:642`) | ~~an unknown child → batch hangs~~ | ✅ **fixed** — keyed by origin, reported under the settled attempt |
| 6 | Delegation edge timeout (`spawn_router.go:1202`) | ~~expires the retired attempt, orphans the successor~~ | ✅ **fixed** — `liveAttempt` targets the running one |
| 7 | Schedule `Forbid` (`sympoziumschedule_controller.go:185`) | ~~finished → fires a second run mid-chain~~ | ✅ **fixed** — successor carries `sympozium.ai/schedule` |
| 8 | Stimulus guard (`sympoziumschedule_controller.go:217`) | ~~finished → same~~ | ✅ **fixed** — carries `sympozium.ai/stimulus` |
| 9 | Web proxy + MCP wait (`openai.go:149`, `mcp.go:270`) | ~~a name that never settles → request hangs to its 10m deadline~~ | ✅ **fixed** — `answersRun` (`proxy.go:44`) follows `status.retryOf` |
| 10 | Request dedupe (`findRecentWebRun`, `openai.go:459`) | ~~terminal → a duplicate request starts a second chain~~ | ✅ **fixed** — carries `sympozium.ai/request-hash`; the successor is the newest non-terminal match |
| 11 | `pipelineInFlight` (`sympoziumschedule_controller.go:472`) | in flight — keys on `sympozium.ai/ensemble`, always carried | ✅ already correct |
| 12 | Sequential + delegation successors (`agentrun_controller.go:1434`) | nothing — guarded by `Phase == Succeeded` | ✅ already correct |
| 13 | `reconcileCompleted` on the retired attempt | done — pod/PVC/RBAC cleanup, finalizer, token budget | ✅ correct; that attempt really did run and spend those tokens |
| 14 | `agentRunsTotal` | a failure with `reason="retried"` | ✅ separable (the `status` attribute is still `failed`) |
| 15 | UI toast (`use-run-notifications.ts:70`) | a failed run → red toast | 🟡 accepted — a failure did occur |
| 16 | Sidebar badge (`sidebar.tsx:89`) | counts toward unseen-failed | 🟡 accepted |
| 17 | Gateway metrics (`apiserver/server.go:4077`) | `ErrorCount` + a request of its own | 🟡 accepted |
| 18 | Feed / canvas / topology chips | red "Failed" (`run-detail.tsx` is the one surface that knows `retried`) | 🟡 accepted |
| 19 | `pruneOldRuns` (`agentrun_controller.go:2136`) | eligible for history pruning, oldest-first | ⚪️ benign — `retryChainTokens` already tolerates a pruned ancestor |
| 20 | `workspacesession.isTerminalPhase` (`:412`) | terminal → releases the workspace lock, successor re-acquires | ⚪️ small window where another session could take the lock |
| 21 | `sympozium workspace` CLI (`workspace_cmd.go:405`) | terminal → drops out of the active list | ⚪️ cosmetic |
| 22 | Ensemble disable path (`ensemble_controller.go:828`) | deletes every run for the persona, chain included | ⚪️ intended |
| 23 | Celln scoped / review-UI paths | n/a — different backend, no gates | ⚪️ n/a |

The through-line: a retired attempt is terminal under a name nothing points at any more, so anything tracking a run **by name** loses the chain. Two mechanisms close that — the parent's `status.delegates[].childRunName` is repointed at the live attempt before the predecessor is retired, and the labels other controllers find live work by (`schedule`, `stimulus`, `request-hash`, `ensemble`) are carried forward. Callers blocked on a name follow `status.retryOf` instead.

kristof-ringleff:
Thanks @AlexsJones I have moved out the comments for 589 and have taken another look at the handling for retried attempts during delegation.

The previous comment is an overview of how a failed status is currently handled throughout. Let me know if you think anything should be classified differently.

AlexsJones (commented):
This is a thorough answer to the delegation question, thank you. Repointing `status.delegates[].childRunName` before the predecessor is retired, walking the lineage in `claimPending` and `RetryChainContains`, and carrying the schedule, stimulus and request-hash labels onto the successor covers every by-name tracker I asked about, and `agentrun_retry_delegate_test.go` pins the retry-then-pass case. Moving the summed-budget docs into #589 settles the other point.

One small thing, not blocking: if `reassignDelegation` fails on a pass where the successor already existed (`created == false`), the verdict resolves as reject but the successor keeps running. It's a double-failure edge and your `discardRetryRun` comment already accepts that cost, so fine to leave.

CI is green at `ce050a64`, including the envtest system suite. Approving on the code. Since this adds `lifecycle.retry` to the API, I'll take one more look at the API shape before it merges.

PR #339, opened 2026-08-18
feat: resume gate

kristof-ringleff:
To be merged after #338

As a continuation of the work in the previous PR this adds the ability for the agent to continue work if the acceptance gate rejected it, ie fix a failed test rather than having to re-implement the whole feature from scratch as currently happens in a retry.

image


AlexsJones (commented):
Thanks for building the resume path on top of #338. Letting the agent fix a failed test in place, instead of redoing the whole task, is a much better loop. This branch carries all of #338's commits and now conflicts with main (525 commits behind), so I'll hold the review until #338 is rebased and settled. When you rebase, basing this branch on the rebased #338 keeps the diff to just the resume gate.

PR #341, opened 2026-08-18
feat: add Databricks AI Gateway provider

### LINKED ISSUE #340: Provider: Databricks
## Goal

Add Databricks as a first-class Sympozium provider so agents can use chat models exposed through Databricks AI Gateway and Model Serving.

Implementation: #341

## Scope

- Add `databricks` to the provider configuration and onboarding flow.
- Support Databricks workspace and AI Gateway base URLs.
- Authenticate securely using an API token or Kubernetes Secret reference.
- Discover available READY chat models from Databricks serving endpoints.
- Exclude embedding-only and unavailable endpoints from model selection.
- Show discovered models in onboarding and AgentRun model dropdowns.
- Keep support for manually entering a model or serving endpoint name.
- Invoke compatible models through the Databricks OpenAI-compatible Chat Completions API.

## Acceptance criteria

- A Sympozium Agent can be configured with `provider: databricks`.
- Databricks credentials remain server-side and are not exposed to the browser.
- Available chat models are fetched dynamically and displayed for selection.
- The selected model can be used successfully in an AgentRun.
- Backend unit tests and the frontend production build pass.

christina-devops:
## Summary
- add Databricks AI Gateway as a provider
- discover READY chat models from Databricks serving endpoints
- expose discovered models in onboarding and AgentRun dropdowns
- support Kubernetes Secret-based Databricks authorization without exposing credentials

## Validation
- `go test ./internal/apiserver`
- `npm run build`
- live HTTP 200 invocation: `databricks-claude-opus-5`
- live HTTP 200 invocation: `databricks-gpt-5-5`

Closes #340

### DIFF
--- a/internal/apiserver/server.go
+++ b/internal/apiserver/server.go
@@ -3796,8 +3796,11 @@ func (s *Server) proxyProviderModels(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	provider := strings.ToLower(strings.TrimSpace(r.URL.Query().Get("provider")))
+
 	// Resolve hostname and check for disallowed IPs.
 	hostname := parsed.Hostname()
+	isDatabricks := provider == "databricks" && isDatabricksHostname(hostname)
 	ips, err := net.LookupHost(hostname)
 	if err != nil {
 		http.Error(w, "cannot resolve baseURL hostname", http.StatusBadRequest)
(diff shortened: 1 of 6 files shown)


AlexsJones (changes requested):
Thanks for this, and for including the live 200s against two real gateway endpoints. The serving-endpoint parsing, the READY gate and the embedding filter all look right, and `TestParseDatabricksServingEndpoints` covers the cases that matter.

Two things I want changed before this goes in, both in `internal/apiserver/server.go`.

1. **The private-IP exemption.** `isDatabricks` switches off the loopback / link-local / private-IP guard for any hostname whose suffix is `databricks.com`, and the request it then makes carries a cluster Secret in the `Authorization` header. A hostname suffix is not evidence about where a name actually resolves, and this endpoint is reachable by anyone who can hit the API (auth is a startup option, see the warning `Handler` logs when the token is empty). If PrivateLink workspaces are the reason you needed it, make it an explicit operator setting rather than something a query parameter turns on.

2. **The secret read.** `?secretName` plus `?namespace` lets a caller name any Secret in any namespace and have its value forwarded as an `Authorization` header to a host of their choosing. `providerAuthorization` will happily pick up `OPENAI_API_KEY` or `API_KEY` from an unrelated secret too. Scope it: read only the secret the selected instance's `authRefs` already points at, or restrict it to a configured namespace.

Smaller: `models := parseProviderModels(body)` runs and is immediately thrown away on the databricks path, so make that an if/else. And nothing ran CI on this one, so `go test ./...` and the web build are unverified here.

PR #344, opened 2026-08-20
fix(agent-runner): skip empty-task seed in prompt-server mode + polymorphic task API

KSiig:
## Summary

Two related fixes for sidecar-driven (prompt-server) mode:

1. **Skip empty-task seed message** — in prompt-server mode, `task=""` is intentional (the sidecar passes prompts via `/ipc/prompts/`). Both providers seeded the message history with an empty-task user message, which Anthropic rejects as HTTP 400 (`"messages: text content blocks must be non-empty"`). OpenAI silently accepted it but wasted a conversation turn.

2. **Accept polymorphic task in CreateRun API** — the CRD already accepts both string and object-form tasks via `TaskSpec`, but the REST API's `CreateRunRequest.Task` was typed as `string`. Sidecars that trigger other agents (e.g. a collector triggering a parser) use the REST API, so they couldn't create sidecar-driven runs.

### Commit 1: Empty-task seed fix

**`cmd/agent-runner/provider_anthropic.go`** and **`cmd/agent-runner/provider_openai.go`**

Guard both the constructor and `ResetContext()`: only seed the task message when the task string is non-empty.

```go
// Before (both providers):
messages: []MessageParam{
    NewUserMessage(NewTextBlock(task)),  // task="" → empty content block
}

// After:
var seed []MessageParam
if task != "" {
    seed = []MessageParam{NewUserMessage(NewTextBlock(task))}
}
```

### Commit 2: Polymorphic task API

**`internal/apiserver/server.go`**

- `CreateRunRequest.Task`: `string` → `json.RawMessage`
- Parse via `TaskSpec.UnmarshalJSON` (existing polymorphic handler)
- Set `&taskSpec` directly on the AgentRun instead of `NewStringTask(req.Task)`
- Backward compatible: callers sending `"task": "do the thing"` still work

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./cmd/agent-runner/ -short` passes
- [x] `go test ./internal/apiserver/ -short` passes
- [x] `gofmt -l` clean
- [x] E2E tested: sidecar-driven runs with `useContext=true` succeed on both Anthropic and OpenAI providers
- [x] E2E tested: REST API accepts both string and object-form task, creates correct AgentRun

### DIFF
--- a/cmd/agent-runner/provider_anthropic.go
+++ b/cmd/agent-runner/provider_anthropic.go
@@ -53,16 +53,21 @@ func newAnthropicProvider(apiKey, baseURL, model, systemPrompt, task string, too
 		anthropicTools = append(anthropicTools, tool)
 	}
 
+	var seed []anthropic.MessageParam
+	if task != "" {
+		seed = []anthropic.MessageParam{
+			anthropic.NewUserMessage(anthropic.NewTextBlock(task)),
+		}
+	}
+
 	return &anthropicProvider{
 		client:      anthropic.NewClient(opts...),
 		model:       model,
 		system:      systemPrompt,
 		initialTask: task,
-		messages: []anthropic.MessageParam{
-			anthropic.NewUserMessage(anthropic.NewTextBlock(task)),
-		},
-		tools:      anthropicTools,
-		toolsBytes: jsonBytes(anthropicTools),
+		messages:    seed,
+		tools:       anthropicTools,
+		toolsBytes:  jsonBytes(anthropicTools),
 	}
 }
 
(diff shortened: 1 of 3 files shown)


AlexsJones (approved):
Good catch on the empty seed message. Anthropic rejecting an empty text block with a 400 while OpenAI quietly burns a turn is exactly the kind of asymmetry that hides in sidecar mode, and guarding both the constructor and `ResetContext` on each provider is the right coverage.

Typing `CreateRunRequest.Task` as `json.RawMessage` and letting `TaskSpec` do the polymorphic decode is the clean fix, since it makes the REST API agree with the CRD instead of maintaining a second shape.

One thing worth tightening, not blocking: the validation moved from `req.Task == ""` to `len(req.Task) == 0`, so a body with `"task": null` is now 4 bytes of RawMessage and sails through into a zero-value TaskSpec. Empty string is deliberate now, fair enough, but `null` looks like an accident rather than a sidecar-driven run. A `bytes.Equal(req.Task, []byte("null"))` check next to the length test would close it.

A test on the object-form task through `createRun` would also be worth having, since that path is the whole point of commit 2 and nothing currently pins it.

Verify is green. Recommending merge, with those two as follow-ups if you would rather not respin.

AlexsJones:
@copilot resolve the merge conflicts in this pull request

AlexsJones (approved):
Both fixes are clean. Seeding the history only when `task` is non-empty stops the empty content block Anthropic rejects, and the `json.RawMessage` task on `CreateRunRequest` keeps string callers working while letting sidecars pass object-form tasks through the polymorphic `TaskSpec`. Backward compatible as far as I can tell. Good to go.

PR #345, opened 2026-08-21
feat(agent-runner): structured output support for Anthropic models

KSiig:
## Summary

- Anthropic `Prompt()` path now builds a synthetic tool from the sidecar's JSON schema and forces `tool_choice` to extract structured JSON via `tool_use.Input`
- `ToolDef.Strict` field passes through to Anthropic and OpenAI tool definitions on the `Chat()` path; local providers silently drop it
- `SupportsStrictSchema()` method added to `LLMProvider` interface
- Azure OpenAI strict mode gated on API version >= 2024-08-01
- Empty seed message skipped in prompt-server mode (fixes `useContext=true` with Anthropic)
- Bedrock `SupportsStrictSchema()` returns false until strict metadata is implemented

## Test plan

- [x] Unit tests pass
- [x] Integration tested: Anthropic structured output via sidecar-driven AgentRun (useContext=true)
- [x] Integration tested: OpenAI structured output via sidecar-driven AgentRun

### DIFF
--- a/cmd/agent-runner/provider.go
+++ b/cmd/agent-runner/provider.go
@@ -108,6 +108,12 @@ type LLMProvider interface {
 	// LLM call succeeded but the model failed to emit valid schema, and
 	// callers receive an error.
 	Prompt(ctx context.Context, prompt string, useContext bool, schema json.RawMessage) (string, []byte, int, int, error)
+
+	// SupportsStrictSchema reports whether the provider's API enforces
+	// schema validation on tool definitions when Strict is set. Providers
+	// that lack grammar-constrained decoding (Ollama, LM Studio,
+	// llama-server) return false and silently ignore the flag.
+	SupportsStrictSchema() bool
 }
 
 // runAgentLoop drives a provider through iterative tool calling until the
--- a/cmd/agent-runner/provider_anthropic.go
+++ b/cmd/agent-runner/provider_anthropic.go
@@ -50,6 +50,9 @@ func newAnthropicProvider(apiKey, baseURL, model, systemPrompt, task string, too
 		}
 		tool := anthropic.ToolUnionParamOfTool(schema, t.Name)
 		tool.OfTool.Description = anthropic.String(t.Description)
+		if t.Strict {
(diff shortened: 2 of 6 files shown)


AlexsJones (changes requested):
Sorry this sat for a month. The core of it is right and still needed: main's Anthropic `Prompt` still sends `ToolChoice: OfNone` unconditionally (`provider_anthropic.go:250`), so a sidecar asking for structured output from Anthropic gets text and hopes it parses. Driving it through a synthetic tool with `tool_choice` forced to that tool is the correct shape, and keeping the text-parse path as the fallback when `parsed == nil` means nothing regresses for providers or calls that do not take that route. The empty-seed fix and the matching `ResetContext` change are a genuine bug fix in their own right, and it is good that `historyText` falls back to the parsed JSON so `useContext=true` does not record an empty assistant turn.

Three things before I take it.

**`SupportsStrictSchema()` has no callers.** It goes on the interface and gets implemented four times, and nothing in the PR or on main ever calls it. Same for `ToolDef.Strict`: the field is added and plumbed into both the Anthropic and OpenAI tool definitions, but nothing anywhere sets it to true, so the OpenAI branch `if t.Strict && supportsStrict` is dead. Either wire a producer in this PR (I assume the intent was the sidecar's tool definitions carrying strict through), or drop both and land the Anthropic structured-output path on its own. I would rather have the second, smaller PR now and the strict plumbing when something actually consumes it.

**The Azure version gate is duplicated.** `newOpenAIProvider` computes `supportsStrict` from the provider name and `AZURE_OPENAI_API_VERSION`, then `SupportsStrictSchema()` computes the same thing again from the same two inputs. Two copies of one rule drift. Pull it into a free function, `openAISupportsStrict(provider string) bool`, and have both call it.

**String comparison on API versions.** `apiVersion >= "2024-08-01"` happens to work for Azure's `YYYY-MM-DD` and `YYYY-MM-DD-preview` forms, since the suffix sorts after the bare date. That is true by luck rather than design, so it wants a one-line comment saying why it is safe, or a comparison on the first ten characters.

Verify is green. Once the dead plumbing question is settled this is good to go.

PR #408, opened 2026-09-01
feat(sessionkey): stable per-conversation session keys

SamuZad:
Add an internal/sessionkey package that derives stable, structured session keys per run source, replacing nanosecond-unique keys:

- chan:<channel>:<chat>[:<thread>] for channel messages, so every message in one Slack thread (or channel, when threading is off) continues one session instead of starting cold
- sched:<name> for scheduled runs, so recurring runs accumulate memory
- web:<instance>:<request-hash> keyed on the proxy idempotency hash, so retries of the same logical request share a session
- mcp:<instance>:<session-id> tied to the actual SSE session
- per-instance web-endpoint keys (previously a literal shared by every Agent, letting instances leak memory to each other)
- adhoc:<random> via crypto/rand for API-server defaults (UnixNano could collide under concurrency)

Hash() gives a 16-char hex digest for DNS-1123-safe derived names and label values. AgentRun.spec.sessionKey stays free-form; nothing here is schema-enforced.

### DIFF
--- a/internal/apiserver/server.go
+++ b/internal/apiserver/server.go
@@ -39,6 +39,7 @@ import (
 	"github.com/sympozium-ai/sympozium/internal/collector"
 	"github.com/sympozium-ai/sympozium/internal/controller"
 	"github.com/sympozium-ai/sympozium/internal/eventbus"
+	"github.com/sympozium-ai/sympozium/internal/sessionkey"
 )
 
 const systemNamespace = "sympozium-system"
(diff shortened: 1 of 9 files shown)


AlexsJones (approved):
Thanks for keeping this focused. I checked that session keys are centralized and stable per channel, schedule, proxy, MCP, endpoint, and sub-agent source, with randomized ad-hoc defaults. Ready for your final look.

PR #409, opened 2026-09-01
feat(workspace): per-session persistent workspaces via WorkspaceSession CRD

SamuZad:
> **Stacked on #408** — the first commit (`feat(sessionkey)`) is shared; this PR is the last 4 commits. Kept as draft until #408 lands, then I'll rebase so only the workspace commits remain.

Gives conversations durable state: consecutive runs in the same session (same Slack thread, schedule, or web session) resume on the same `/workspace` files instead of starting from an empty `emptyDir`.

## What's in here

- **`WorkspaceSession` CRD** — owns one RWO PVC per `(agent, sessionKey)`. Spec: `agentRef`, `sessionKey`, `size`, `storageClassName`, `idleTTL`. Owned by its Agent and owning its PVC, so deletion cascades via GC. Deterministic hash-based naming (`ws-<agent>-<hash>`, PVC `...-g<generation>`) keeps raw session keys out of object names.
- **Opt-in** via `workspace.perSessionPVC` on `Agent.spec`, `Ensemble.spec`, or per-persona (persona replaces the ensemble value). Sub-agents stay ephemeral by design.
- **Session locking with FIFO admission** — RWO PVCs cannot multi-attach, so only one AgentRun per session may hold the workspace: Running/Serving peers block; between Pending waiters only the older run blocks (CreationTimestamp, then Name), so waiters form a strict FIFO and can never mutually deadlock. Blocked runs surface a `Blocked/SessionBusy` condition and requeue without creating a Job.
- **Idle reclamation** — hourly sweeper deletes sessions untouched for `idleTTL` (default 30d; `0s` disables) when no live run holds them.
- **Grow-only online resize** gated on the StorageClass's `allowVolumeExpansion`; refused shrinks/expansions surface as a `ResizeBlocked` condition rather than mutating the PVC.
- **`workspace-marker` init container** records run lineage in `/workspace/.sympozium/state.json` (previous marker nested under `previousRun`). Implemented as an `agent-runner workspace-marker` subcommand because the image is distroless — no shell.
- **CLI**: `sympozium workspace list|show|delete|exec` (exec spins up a transient debug pod mounting the PVC and prints the kubectl commands; refuses while a live run holds the RWO volume).

## Test notes

- New controller tests (`workspacesession_controller_test.go`, incl. a 9-case FIFO admission table) and CLI tests (`workspace_cmd_test.go`).
- `convergenceFixture` in `ensemble_parity_test.go` now populates `Workspace` so the ensemble-expressibility parity test covers the new field.
- `go test -short ./...` is green on top of `c3bab88`.


AlexsJones (approved):
Thanks for keeping this focused. I checked that per-session workspaces have explicit CRD lifecycle, RWO serialization, idle reclamation, PVC recreation state, RBAC, generated manifests, and controller coverage. Ready for your final look.

PR #414, opened 2026-09-02
feat(harness): close critical GA security gaps

AlexsJones:
## Summary

- keep remote MCP credentials outside the external adapter trust boundary by rejecting harness AgentRuns with `Agent.spec.mcpServers`
- harden the mediated SkillPack MCP endpoint with JSON-RPC ID compatibility, input-schema validation, request/output/concurrency bounds, HTTP timeouts, and Origin checks
- run the persistent lifecycle suite daily against both maintained Pi and Hermes adapters using a pinned in-cluster Ollama model
- document the credential boundary and scheduled integration path

Closes the three critical work streams identified in #349. The scheduled workflow still needs to pass on this branch before the real-adapter checkbox is marked complete.

## Validation

- `go test ./...`
- `helm lint charts/sympozium`
- workflow YAML parse
- `git diff --check`

### DIFF
--- a/.github/workflows/integration-kind.yaml
+++ b/.github/workflows/integration-kind.yaml
@@ -114,7 +114,7 @@ jobs:
                         cpu: 500m
                         memory: 1Gi
                       limits:
-                        memory: 2Gi
+                        memory: 3Gi
                     volumeMounts:
                       - name: models
                         mountPath: /root/.ollama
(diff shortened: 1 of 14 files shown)


AlexsJones (commented):
Thanks for the thorough security pass. One request-framing gap remains: the JSON decoder accepts a valid request and ignores trailing JSON. After the first decode, decode once more and require `io.EOF`, then add a regression with two concatenated values. That keeps the one-request-per-HTTP-body boundary explicit.

PR #419, opened 2026-09-03
docs: lead README with family stack GIF

AlexsJones:
## Summary
- README hero now uses `docs/assets/family-stack.gif` (family stack positioning) instead of the topology/dashboard demo.
- `demo.gif` kept lower under a short **What it looks like** section after Quick Install.

## Test plan
- [ ] Confirm GitHub README renders the stack GIF at the top
- [ ] Confirm `demo.gif` still appears under What it looks like
- [ ] Spot-check mobile README width (~800)


AlexsJones (commented):
The family-stack visual is a strong lead, but this change makes the README load both the new 5.1 MB hero and the existing 7.0 MB demo, roughly 12 MB total. Please optimize the hero to a much smaller asset or avoid embedding both animations on the main README so the landing page stays usable on slower links.

PR #420, opened 2026-09-04
fix(helm): suspend postgres MCP server by default

### LINKED ISSUE #397: Postgres MCP server ships without DATABASE_URI and crashloops on every fresh install
## Summary

The chart deploys a `postgres` MCPServer (crystaldba/postgres-mcp) with no
`DATABASE_URI`, so it crashloops from the first minute after a clean install.
Worse, it reports `READY: true` while doing so.

## Environment

- sympozium v0.10.51, chart sympozium-0.10.51
- Fresh install on a cleaned cluster

## Actual behaviour

Two minutes after `sympozium install`:

$ kubectl logs -n sympozium-system deploy/postgres --tail=5
ValueError: Error: No database URL provided. Please specify via
'DATABASE_URI' environment variable or command-line argument.
stdio process exited: exit status 1, restarting with backoff

The Deployment's env block contains only stdio-adapter variables
(`MCP_STDIO_ADAPTER`, `STDIO_CMD`, `STDIO_ARGS`, `STDIO_SERVER_NAME`,
`STDIO_PORT`) — no `DATABASE_URI`, no `envFrom`, and no matching Secret in
the namespace.

On a previous 9-day-old install this pod had accumulated 55 restarts.

## Secondary issue: READY is misleading

NAME       TRANSPORT   READY   SUSPENDED   TOOLS
postgres   stdio       true                

The liveness probe targets `/healthz` on port 8080, which is served by the
stdio adapter, not by `postgres-mcp`. The adapter stays up while its child
process dies in a loop, so the failure is invisible in
`kubectl get mcpservers`. An agent granted this MCP server would see it as
healthy and only fail at tool-call time.

## Suggested fix

Either don't deploy MCP servers that require configuration by default, or
create them with `suspended: true` so the operator opts in. Separately, the
readiness signal should reflect the state of the wrapped process.

## Workaround

kubectl patch mcpserver postgres -n sympozium-system \
  --type=merge -p '{"spec":{"suspended":true}}'

Also worth noting: `postgres-mcp` starts in UNRESTRICTED mode, which would
give agents unrestricted query access to whatever database is eventually
wired in.

Ultronen:
## Summary

- suspend the default PostgreSQL MCPServer until its operator-provided `DATABASE_URI` is configured
- add a regression test that loads and renders the embedded Helm chart and verifies the PostgreSQL MCPServer is suspended

After configuring the `mcp-postgres-token` secret, operators can explicitly set `spec.suspended` to `false` to start the server.

## Testing

- `make fmt`
- `make vet`
- `make test`
- `go build ./...`
- `helm lint charts/sympozium/`
- rendered `templates/default-mcp-servers.yaml` and verified `postgres.spec.suspended: true`

## AI assistance

This contribution was developed with AI assistance. The final diff was independently reviewed and all listed checks were run locally.

Fixes #397

### DIFF
--- a/charts/sympozium/files/mcp-servers/postgres.yaml
+++ b/charts/sympozium/files/mcp-servers/postgres.yaml
@@ -8,6 +8,8 @@ metadata:
     sympozium.ai/catalog: "true"
 spec:
   transportType: stdio
+  # DATABASE_URI is operator-provided; do not start the server until it is configured.
+  suspended: true
   toolsPrefix: pg
   timeout: 60
   toolsDeny:
--- a/internal/helmchart/helmchart_test.go
+++ b/internal/helmchart/helmchart_test.go
@@ -1,0 +1,63 @@
+package helmchart
+
+import (
+	"io"
+	"strings"
+	"testing"
+
+	"helm.sh/helm/v3/pkg/chartutil"
+	"helm.sh/helm/v3/pkg/engine"
+	utilyaml "k8s.io/apimachinery/pkg/util/yaml"
+)
+
+func TestDefaultPostgresMCPServerStartsSuspended(t *testing.T) {
+	ch, err := Load()
+	if err != nil {
+		t.Fatalf("Load() error = %v", err)
+	}
+
+	values, err := chartutil.ToRenderValues(ch, map[string]any{}, chartutil.ReleaseOptions{
+		Name:      "sympozium",
+		Namespace: "test-system",
+		IsInstall: true,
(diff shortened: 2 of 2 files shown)


AlexsJones (approved):
Thanks for making the default chart safe before credentials exist. The PostgreSQL MCPServer now starts suspended, and the chart-level test renders the embedded chart, locates the exact object, and fails unless `spec.suspended` is true. This directly covers the crashloop path without changing configured opt-in behavior. Ready for your final look.

PR #427, opened 2026-09-06
fix(celln): restrict router ingress to controller namespace and pods (M0)

AlexsJones:
## M0 first increment

Part of #426; partial mitigation for #331, not closure.

- Add mandatory Celln router ingress NetworkPolicy in `celln-system`, independent of the optional general agent policies.
- Allow only TCP 8788 from the actual controller pod selector AND configured control-plane namespace; copied labels in a tenant namespace do not qualify.
- Render the real Helm chart in tests and check selector semantics for defaults, custom namespace, general policies disabled, and Celln disabled.
- Document that this needs an enforcing CNI and does not substitute for inbound authentication. The API server is deliberately not given execution access; its current TCP-only capability probe may report unavailable until M0's least-privilege readiness path lands.

## Validation

- `go test -race ./internal/controller ./api/v1alpha1 -count=1` passed, including the four Helm render/selector cases.
- `helm lint charts/sympozium` passed.
- No cluster changes and no cross-namespace CNI enforcement proof claimed; that remains in M0's deployed E2E gate.

## Dependencies and limits

Companion Celln PR adds distinct inbound credentials and cancellation forwarding. The existing chart command/image versions, credential distribution and TLS topology still require a coordinated update. This draft is a reviewable first increment, not the completed M0 deployment. #331 and #426 stay open.

### DIFF
--- a/charts/sympozium/templates/celln-network-policy.yaml
+++ b/charts/sympozium/templates/celln-network-policy.yaml
@@ -1,0 +1,31 @@
+{{- if .Values.celln.enabled }}
+# Mandatory for this privileged execution ingress, independently of the
+# optional policies for ordinary agent pods. Requires an enforcing CNI.
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: celln-router-ingress
+  namespace: celln-system
+  labels:
+    app.kubernetes.io/part-of: sympozium
+spec:
+  podSelector:
+    matchLabels:
+      app.kubernetes.io/name: celln-router
+  policyTypes:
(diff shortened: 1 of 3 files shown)


AlexsJones (commented):
Thanks for keeping namespace and pod selectors in the same ingress peer. The chart scopes TCP 8788 to the controller namespace and labels, and the render tests cover copied tenant labels and disabled general policies. This is a sound policy slice, but it remains a coordinated draft with Celln #70: the policy does not authenticate callers or prove CNI enforcement. Keep #331 open until the compatible deployed path is qualified.

PR #428, opened 2026-09-07
feat(celln): bind experimental in-cell Harness runs with frozen authority

AlexsJones:
Advances #426; paired with Celln feat/harness-dispatch-binding (stacked on celln#73).

Adds disabled-by-default spec.celln.harness advanced AgentRun API for the native reference-functions adapter. The controller freezes task, selected model, runtime/closure, model grant and borrowed tools into celln.dev/v1alpha2. Retry preserves frozen identities; mismatched receipt versions refuse. Unsupported provider/credential options refuse instead of silently falling back. Existing container-replacing task.mode=harness and AgentRuntime OCI contracts remain separate.

Actual proof: built controller + isolated Kind API + authenticated host dispatcher + warm-forked KVM cell + DeepSeek + two separate tool binaries. add(37,5) -> 42, multiply(42,2) -> 84, final 84; matching receipt/audit, no Job, zero live cells after dissolution. Namespace and temporary credential copy removed. Evidence is in paired Celln docs/evidence/harness-dispatch-controller-2026-09-07.json; reproducible hook and guide added here.

Validation: generated CRDs/deepcopy, go build ./..., controller/API/webhook/orchestrator tests including race detector; actual controller/model hook passed. CELLN_HARNESS_ENABLED=true required explicitly.

Not the finished UX or deployed router/DaemonSet proof; not general OCI/Pi/Hermes compatibility, per-tool attested receipts, fleet replay safety or conversational lifecycle. Keep epic open.


AlexsJones (commented):
Thanks for keeping the reference Harness behind an explicit gate. The controller freezes model/task/tool authority, rejects unsupported provider options, and checks receipt version against the frozen request. The direct-controller proof is a useful scoped result. Keep this draft paired with Celln #74/#76 until the dependency stack is reviewed; it does not establish general HarnessSession compatibility or per-tool attested receipts.

PR #429, opened 2026-09-07
fix(celln): wire durable router deployment with explicit trust configuration

AlexsJones:
Part of #426; stacked on #427. Requires Celln PRs #70 and #75 in the pinned router image.

Replaces per-node, host-binary-mounted routers with unprivileged image-digest-pinned Deployment replicas using identical explicit dispatcher endpoints, distinct client/backend Secret mounts, and an operator-managed shared ownership PVC. Makes Celln and privileged legacy host installation separate opt-ins. Rejects missing trust/storage/image configuration and requires explicit acknowledgement of plaintext transport. Adds migration, rotation, ownership-retention and rollback limits.

Validation: helm lint charts/sympozium; go test ./internal/controller -count=1; git diff --check. Render tests exercise replica wiring, no host/ephemeral router volumes, credential rotation-compatible mounts, independent installer opt-in, negative configuration cases and existing ingress selector semantics.

Not ready for merge as completed M0: actual shared filesystem qualification, deployed two-node KVM path, TLS, authenticated eligibility, and coordinated host installer upgrade/rollback remain open. No claim that chart rendering proves deployed execution.


AlexsJones (commented):
Thanks for replacing host-mounted router binaries with pinned unprivileged replicas. The credential mounts allow rotation and the ownership PVC is explicit. One release decision needs to stay visible: this changes Celln from enabled by default to disabled and requires new credential/storage/image configuration. Please confirm that upgrade behavior and its migration notice before release; I recommend the explicit opt-in, with a tested upgrade/rollback sequence for installations that already enabled Celln.

PR #430, opened 2026-09-07
feat(celln): authenticated capability status and scoped discovery credentials

AlexsJones:
Advances M0 in #426. Stacked on #429; runtime dependency sympozium-ai/celln#80.

Replaces TCP-only API availability with authenticated/versioned GET /v1/capabilities. Requires explicit enablement and read-only credentials, refuses redirects and inconsistent/invalid reports, and checks node preflight/capacity. No fallback to execution credentials or public health. UI qualifies positive status as node preflight, not selected Harness/artifact readiness.

Chart supplies distinct discovery Secret references and namespace AND API-Pod-scoped ingress. API server receives no execution/backend credential. Existing enabled installations require new discovery Secrets and compatible images.

Validation: full API/controller tests, race-enabled capability tests, Helm render/security tests and lint, web build. Existing dependency install reports 17 advisories; versions unchanged.

Deployed proof PASSED: actual API -> router Service -> two dispatchers; read-only submit/poll/audit/cancel all 401; copied API labels in a tenant namespace blocked; unlabelled control-plane Pod blocked and allowed after matching label; credential mismatch fails unavailable and restoration recovers without API restart. Evidence 83135b9: docs/evidence/celln-capability-discovery-2026-09-07.md.

Important qualification: test image needed explicit non-root UID; absent NATS defeated default liveness, so test used a bounded startupProbe and explicit API-only flag. #431 tracks that discovered startup/UI flag issue. Not a full Helm release, browser E2E or Harness artifact-readiness proof. Full epic remains open.


AlexsJones (commented):
Thanks for replacing TCP reachability with an authenticated contract. The probe disables redirects and proxy inheritance, bounds credentials/responses, and checks each eligible node against the reported count; the API pod receives only the discovery credential. The UI correctly limits the claim to node preflight. Keep the draft tied to Celln #80 and the coordinated release gate, and carry #432's normal-probe startup evidence into the final deployment record so the earlier startupProbe workaround is not mistaken for the shipped configuration.

PR #432, opened 2026-09-07
fix(apiserver): bound optional NATS startup and honor disabled UI

AlexsJones:
Addresses #431; advances #426. Stacked on #430.

Empty event-bus URL skips NATS. API uses a five-second initial provisioning context, closes failed connections, and starts HTTP without streaming. Initial failure requires restart after NATS restoration; successfully initialized buses preserve background reconnect. Reconnect initialization is synchronized. Chart explicitly honors --serve-ui=false.

Validation: race-enabled API, eventbus and controller suites; bounded/cancelled/non-responsive/unreachable startup tests; chart flags and Helm lint. Deployed API with normal probes and zero restarts: disabled NATS starts HTTP immediately; unreachable NATS starts after five seconds. Authenticated Celln discovery remains functional; disabled UI root is 404.

Real NATS 2.11.4 retained-store process-restart regression passed five cycles under race detector. Startup context is cancelled after successful initialization; identified events arrive before and after SIGKILL/restart on the original connection and subscription. Reproducible opt-in test committed at 046e82d. No repository dependency version changes.

Evidence: docs/evidence/apiserver-nats-startup-2026-09-07.md. Limits: not broker data-loss/replicated failover qualification, new API streaming/browser E2E, or full epic completion. Awaiting stacked review/merge.


AlexsJones (commented):
Thanks for separating initial NATS startup from established-connection reconnects. Empty URLs return a nil interface, initial provisioning uses the five-second context, failed initialization closes the connection, and the chart explicitly disables the UI when requested. The committed reconnect test checks the original connection/subscription after cancelling the startup context. This looks coherent within #431's scope; retain draft status until the #430 stack is ready and the recorded deployed evidence is checked at the release revision.

PR #433, opened 2026-09-07
docs(celln): define runtime placement and approved borrowed-tool contract

AlexsJones:
Advances M1 of #426 and the refreshed #335. Stacked on #428.

ADR separates approved runtime identity, runtime/tool placement and explicit lending. Defines a separately gated AgentRuntime Celln profile; immutable executable-tool catalogue and user submission versus operator approval; exact distribution/prewarming; signed precomposition compatible with the current dispatcher; frozen UIDs/revisions and shrinking authority; and external conversation state with disposable bounded cells per turn.

Preserves OCI and existing HarnessSession semantics, treats informational conformance as insufficient authority, and keeps MCP/SkillPacks distinct. Specifies ordered implementation and conformance gates through the full user journey.

This is an implementation design, not new CRDs/UI functionality or a completion claim. M0 release/recovery/retention gates and M1-M4 implementation remain open. Validation: checked against current runtime, AgentRun and HarnessSession types, reference Harness guide and Celln requirements; git diff --check passes. No runtime tests or cluster mutations for this documentation-only change.


AlexsJones (commented):
Thanks for separating runtime identity, placement and tool authority in one contract. The proposed external checkpoint plus disposable bounded turn model avoids implicitly turning one-shot cells into persistent services. Please confirm that lifecycle choice before the session API implementation: I recommend it, with the stated single-turn lease, frozen revisions, current-authority recheck and no replay of ambiguous effects. Catalogue metadata and the reference arithmetic proof should remain separate from a supported full-Harness claim.

PR #434, opened 2026-09-07
feat(celln): immutable tool catalogue and unprivileged submissions

AlexsJones:
First M1 catalogue schema/RBAC slice for #335 and #426. Stacked on #433.

Adds namespace-scoped CellnToolSubmission and operator-managed CellnTool metadata with immutable revisions, bounded artifact/publisher/schema references, canonical entry point, invocation ABI, lane/platform and resource ceilings. Current profile explicitly refuses nonempty inputs/egress and non-none workspace. No default Ready/Approved state; metadata creation grants no execution authority.

Chart adds unbound submitter/reviewer/verifier roles with no automatic grants to existing workloads. Generated deepcopy and CRDs are synchronized across both charts. A 71-character bound on existing immutable references matches their existing exact hash pattern.

Actual isolated Kubernetes v1.35 proof passes: forged submission status stripped, user catalogue/status writes forbidden, cross-namespace submissions forbidden, reviewer/verifier duties separated, immutable spec and malformed hashes/paths/images/limits rejected. Script uses explicit kind-celln-deployed kubeconfig, dummy metadata and cleans only its test namespace/roles; new CRDs remain installed. No model, cell or Job execution.

Validation: make generate; race-enabled API/controller/webhook/API-server suites; both chart lints; repeated real schema/RBAC test; git diff --check. Guide: docs/guides/celln-tool-catalogue.md.

Not implemented here: signature/schema-document verification, operator workflow controllers, runtime profile, distribution/prewarming, authority resolver, lending UI or dispatch linkage. Those remain next steps; this does not complete #335 or the epic.


AlexsJones (commented):
Thanks for making user submission separate from operator publication. The source schema makes revisions immutable, bounds hashes/paths/limits, refuses unsupported inputs/egress/workspaces, and the roles are unbound with separate publication and status authority. The integration script exercises those denials without executing a tool. Keep this marked as metadata-only pending the resolver/distribution slices; the next trust-boundary regression should delete and recreate a same-name tool and prove a frozen UID/revision cannot retarget it.

PR #440, opened 2026-09-07
feat: add tolerations for agent pods

SamuZad:
This PR adds support for tolerations to `Agent` config, `AgentRun` spec, and `Ensemble` personas. This in turn allows us to reliably schedule AgentRun pods to dedicated node groups with taints.


AlexsJones (approved):
Thanks for carrying tolerations through the run-creation paths. I checked the Agent/Ensemble/AgentRun fields, generated schema and deepcopy changes, inheritance paths, and both job and server pod templates. The added job tests cover configured and absent tolerations, and the ensemble parity fixture includes the field. Verify is green. Ready for your final look.

PR #466, opened 2026-09-09
Fix #415: Gate sequential edges for subagent batch children

### LINKED ISSUE #415: [BUG] Subagent spawning: sequential edges trigger independently before parent collates batch results
## Description

When an Ensemble uses `spawn_subagents` to fan out work across multiple subagents, each child subagent may **incorrectly trigger its own sequential edges** before the parent agent has finished collecting all batch results. This creates race conditions and breaks ensemble orchestration.

### Expected Behavior
1. Parent agent spawns N subagents via `spawn_subagents` with strategy="parallel" or strategy="sequential"
2. **All N subagents complete independently** and report results back to parent via IPC result files
3. Parent agent **collates all batch results** and publishes a single aggregated batch result
4. **Only after collation is complete** do downstream sequential edges (if any) execute
5. Downstream stages receive parent-collated results, not individual subagent outputs

### Actual Behavior
1. Parent agent spawns N subagents
2. Individual subagents complete
3. **Each subagent independently triggers its own sequential edges** (if configured)
4. Multiple downstream executions occur prematurely/in parallel
5. Orchestration sequence is broken; parent has no collation point

### Impact
- **High** — Any Ensemble using fan-out via `spawn_subagents` in any stage which is not the last of the workflow  is affected
- Sequential workflow guarantees are broken
- Downstream stages may execute before upstream collation completes
- Potential for data races and false success/failure states

### Root Cause
The spawn router logic (internal/controller/spawn_router.go) may not properly gate downstream sequential relationships on batch result publication. Individual child completion handlers may be independently triggering sequential edges instead of deferring to batch-level orchestration.

### Reproduction Steps

1. Create an Ensemble with:
   - An agent that calls `spawn_subagents(strategy="parallel", tasks=[...])` .
   - That Agent must be an intermediate stage in a workflow and must have a sequential edge . For my example I had an agent running through 100s of repos and based on the collected output , the next agent was to take action.
   - Multiple child agents spawned by that agent 
   - A downstream agent in a sequential relationship to the lead agent

2. Run the ensemble and monitor logs for child completion

3. Observe: Downstream agent may start before `spawn_subagents` tool in lead agent returns

4.  From TUI , you will see multiple instances of the downstream agent being created , where in practice it should have been 1

### Probable Related Code Files 
- `sympozium/internal/controller/spawn_router.go` (line 668: comment about parent's blocking spawn_subagents tool hanging)
- `sympozium/internal/controller/spawn_router_subagent_test.go` (TestSubagentBatch_SequentialChildrenSettle — test exists but may not fully cover root cause)
- `sympozium/internal/controller/agentrun_controller.go` (sequential edge triggering logic)
- `sympozium/internal/ipc/bridge.go` (batch result delivery)

### Test Coverage Gap
The test `TestSubagentBatch_SequentialChildrenSettle` validates that children "settle" before batch results are published, but does NOT validate that sequential edges wait for batch publication — only that individual task settlement works.

abindg:
## Summary
This PR fixes issue #415 where subagent children spawned by spawn_subagents could independently trigger downstream sequential edges before the parent completed collation.

## Root Cause
Sequential successor triggering ran on completed runs without distinguishing:

- parent orchestration runs
- subagent batch child runs

So each child completion could enter the sequential trigger path and fan out downstream runs prematurely.

## Changes
1. Tag batch-spawned child runs with sympozium.ai/subagent-batch-id during spawn.
2. In sequential successor triggering, skip runs carrying that batch label.
3. Keep sequential orchestration owned by the parent run only.

## Test
Add regression test coverage for:
1. child batch label propagation
2. child runs not triggering sequential successors
3. parent run triggering exactly one sequential successor

## Why this resolves the issue ?
The sequential handoff now occurs only from the parent orchestration point, after batch collation, preventing duplicate or premature downstream stage execution.

### DIFF
--- a/internal/controller/agentrun_controller.go
+++ b/internal/controller/agentrun_controller.go
@@ -1377,6 +1377,12 @@ func (r *AgentRunReconciler) triggerSequentialSuccessors(ctx context.Context, lo
 	if agentRun.Spec.AgentRef == "" {
 		return nil
 	}
+	// Child runs spawned by spawn_subagents batches must not independently
+	// execute sequential edges. The parent run collates batch results first,
+	// then it is the only run that should trigger downstream sequential stages.
+	if agentRun.Labels["sympozium.ai/subagent-batch-id"] != "" {
+		return nil
+	}
 	var sourceInst sympoziumv1alpha1.Agent
 	if err := r.Get(ctx, types.NamespacedName{Name: agentRun.Spec.AgentRef, Namespace: agentRun.Namespace}, &sourceInst); err != nil {
 		return nil // Instance gone — skip.
--- a/internal/controller/agentrun_delegation_executor_test.go
+++ b/internal/controller/agentrun_delegation_executor_test.go
@@ -247,7 +247,7 @@ func TestTriggerSequentialSuccessors_SubagentChildrenDoNotTrigger(t *testing.T) 
 	ensemble := &sympoziumv1alpha1.Ensemble{
 		ObjectMeta: metav1.ObjectMeta{Name: "team", Namespace: "default"},
 		Spec: sympoziumv1alpha1.EnsembleSpec{
-			AgentConfigs: []sympoziumv1alpha1.AgentConfigSpec{{Name: "reviewer", Schedule: &sympoziumv1alpha1.ScheduleSpec{Task: "review"}}},
+			AgentConfigs: []sympoziumv1alpha1.AgentConfigSpec{{Name: "reviewer", Schedule: &sympoziumv1alpha1.AgentConfigSchedule{Task: "review"}}},
 			Relationships: []sympoziumv1alpha1.AgentConfigRelationship{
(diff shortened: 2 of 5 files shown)


AlexsJones (changes requested):
Thanks for adding the child gate and parent-only successor regression. Please avoid putting the raw BatchID into a Kubernetes label: the existing naming path sanitizes and bounds arbitrary IDs, but the new label uses the original string. An ID containing a slash or longer than 63 characters now makes child creation fail even though its generated run name is valid. Use a bounded safe label token (and an annotation if the full ID is needed), with a regression covering those IDs.

AlexsJones (approved):
Thanks for the bounded label token plus annotation. Using `batchNameToken` for the label and keeping the full ID in an annotation fixes the validation failure, and the slash and long-ID tests cover it. The child gate in `triggerSequentialSuccessors` is the right behavior. Ready for your final look.

PR #469, opened 2026-09-09
Fix #468: Propagate traceparent on LLM provider calls

### LINKED ISSUE #468: feature (Observability) - Sympozium doesn't propagate AgentRun trace/run ID as a header on LLM/MCP calls — hard to correlate runs with downstream gateway traces
## Description
Each AgentRun gets a unique trace context (otel.dev/traceparent annotation, injected as TRACEPARENT env var), but this is not propagated on outbound LLM provider calls. If a team routes LLM traffic through any observability-aware gateway or proxy (e.g. Agentgateway, or similar), there's no reliable way to tie a specific AgentRun to the corresponding downstream request/span.

## Root Cause
`cmd/agent-runner/provider_openai.go` and `cmd/agent-runner/provider_anthropic.go` build their LLM clients with plain HTTP transports — `TRACEPARENT` is never attached to these requests.
By contrast, cmd/agent-runner/memory_tools.go wraps its client with `otelhttp.NewTransport`, and i`nternal/mcpbridge/client.go` explicitly injects trace context via the OTel propagator. LLM calls are the one path that's missing this.

## Current Workaround and Its Limits
The only correlation mechanism today is providerHeaders (e.g. a custom run-id header), sent as a static string. It's a plain map[string]string with no templating support, and every AgentRun-creation path (`agent_controller.go`, `channel_router.go`, `stimulus.go`, `sympoziumschedule_controller.go`, `apiserver/server.go's createRun`, the TUI's `tuiCreateRun)` copies it verbatim from the Agent spec.

## That means:

- A unique value can only be set today by hand-crafting the AgentRun object directly (bypassing TUI, REST create-run, and schedules).
- TUI-triggered and Schedule-triggered runs always inherit the same static Agent-level value — there's no per-run override available to them.
- The practical fallback is approximating correlation by time window (matching a run's start/completion timestamps against gateway trace timestamps), which breaks as soon as two runs of the same Agent overlap.

## Proposed Fix
1. Wrap the LLM provider HTTP transports with otelhttp.NewTransport (same pattern already used for memory-server calls), so traceparent is sent on every LLM call.
2.  Support a run-scoped placeholder in providerHeaders (e.g. resolved to the AgentRun name or a generated run ID) so a header template set once at the Agent/Ensemble level resolves to a unique value per generated AgentRun — consistently across CLI, TUI, schedule, channel, and web-proxy creation paths.

## Acceptance Criteria
- LLM provider calls carry a traceparent header derived from the AgentRun's trace context, so a trace-aware downstream gateway automatically joins the same trace_id without any extra configuration.
- (Optional/fallback) providerHeaders supports a run-scoped placeholder for gateways that cannot join traces via traceparent, resolving to a unique value per generated AgentRun across all AgentRun-creation call sites.
- Regression test proving two runs of the same Agent/persona produce distinct trace_ids on their outbound LLM calls.

abindg:
## Description

This PR fixes issue #468: LLM provider calls (OpenAI-compatible and Anthropic) carried no trace context at all, making it impossible to correlate a specific AgentRun with corresponding downstream spans on a trace-aware gateway or proxy.

## Root Cause
**provider_openai.go**  and **provider_anthropic.go** built their LLM SDK clients with plain HTTP transports. Unlike the memory-server call path in memory_tools.go, which already wraps its client in otelhttp.NewTransport, the LLM call path never attached a traceparent header.

## Changes

- Added tracingHTTPClient() in **observability.go** — an http.Client wrapping otelhttp.NewTransport(http.DefaultTransport), matching the existing memory-server pattern.
- Wired it into newOpenAIProvider via openaioption.WithHTTPClient(...), covering OpenAI, Azure OpenAI, Ollama, LM Studio, and any other OpenAI-compatible backend.
- Wired it into newAnthropicProvider via anthropicoption.WithHTTPClient(...).
- Added provider_traceparent_test.go with two regression tests, TestCallOpenAI_PropagatesTraceparent and TestCallAnthropic_PropagatesTraceparent, each seeding a parent span context and asserting the outbound LLM call's traceparent header carries the correct trace ID.

## Why this resolves the issue
Every LLM call now automatically carries the AgentRun's real trace context. A trace-aware downstream gateway or proxy with clientSampling enabled joins the same trace_id natively, without requiring any manual per-run header configuration.

Issue
Closes #468

### DIFF
--- a/cmd/agent-runner/observability.go
+++ b/cmd/agent-runner/observability.go
@@ -18,6 +18,7 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/metric"
+	"go.opentelemetry.io/otel/propagation"
 	"go.opentelemetry.io/otel/trace"
 )
 
@@ -255,13 +256,15 @@ func traceMetadata(ctx context.Context) map[string]string {
 	}
 }
 
-// tracingHTTPClient returns an *http.Client whose transport injects the W3C
-// traceparent header (via the global OTel propagator) into every outbound
-// request, so LLM provider calls join the same trace as the rest of the run
-// instead of arriving with no trace context at all. Mirrors the pattern
-// already used for memory-server calls (see memoryHTTPClient).
+// tracingHTTPClient returns an *http.Client that injects traceparent into
+// every request. Propagator is pinned, not left to the global, since
+// initObservability skips telemetry.Init (which sets the global) when OTel
+// is disabled or the collector is unreachable.
 func tracingHTTPClient() *http.Client {
-	return &http.Client{Transport: otelhttp.NewTransport(http.DefaultTransport)}
+	return &http.Client{Transport: otelhttp.NewTransport(
+		http.DefaultTransport,
+		otelhttp.WithPropagators(propagation.TraceContext{}),
+	)}
 }
 
 func formatTraceparent(sc trace.SpanContext) string {
--- a/cmd/agent-runner/observability.go
+++ b/cmd/agent-runner/observability.go
@@ -6,12 +6,14 @@ import (
 	"fmt"
 	"log"
 	"net"
+	"net/http"
 	"os"
 	"path/filepath"
 	"strings"
 	"time"
 
 	"github.com/sympozium-ai/sympozium/pkg/telemetry"
+	"go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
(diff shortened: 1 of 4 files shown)


AlexsJones (changes requested):
Thanks for wiring both provider transports. Please make propagation work when the collector is disabled or unreachable: `initObservability` returns before `telemetry.Init` registers the global propagator, while this new transport relies on that global. The runner already extracts TRACEPARENT explicitly, but the default no-op propagator still drops it outbound. Set the transport's W3C propagator explicitly or initialize propagation independently of export, and test both providers with no SDK/exporter and two different run contexts. The current tests install the global propagator themselves, hiding this path.

AlexsJones (commented):
Reviewed. Clean, well-scoped fix that correctly resolves #468 — build, `go vet`, `gofmt`, and `-race` all pass, including the new tests.

What's good:
- `tracingHTTPClient()` correctly pins `propagation.TraceContext{}`, making injection deterministic and independent of global state (which really is no-op when `initObservability` skips `telemetry.Init`).
- The regression tests are meaningful. They pass because otel v1.40's `noopTracer.Start` carries forward the `nonRecordingSpan` seeded by `trace.ContextWithSpanContext`, so propagation is validated with no global propagator registered — nice.

Minor, non-blocking:
1. `tracingHTTPClient()` allocates a new client on every provider construction (once per run). Fine at this scale, but a package-level var (like `memoryHTTPClient`) would be more consistent.
2. Propagator inconsistency with the memory path: `memory_tools.go` relies on the global propagator while the LLM path pins `TraceContext{}`. Identical today (global is also `TraceContext`), but they'd silently diverge if baggage is ever added. Worth a comment or reconciling both.
3. `TestTracingHTTPClient_IgnoresGlobalPropagator` — the self-check constructs a fresh empty `CompositeTextMapPropagator` and asserts it injects nothing, rather than reading `otel.GetTextMapPropagator()`. It passes regardless, but the guard verifies something other than what its name/comment claim.

None of these block merge.

PR #484, opened 2026-09-11
fix(eventbus): survive NATS being unavailable at startup and consumer reaping

SamuZad:
Builds on #254, which made subscriptions self-heal after a NATS restart. Three gaps remained:

- **Consumers were reaped during any fetch gap.** Ephemeral subscribe consumers used the server default `InactiveThreshold` (5s). A NATS rolling restart, a network blip, or a handler blocked on a full channel reaped the consumer, and the `DeliverNew` recreate then silently dropped every event published in between. Set it to 1h so the consumer and its cursor survive the gap. Trade-off: consumers orphaned by a recreate now linger up to 1h before the server reclaims them (one per recreate, so bounded).
- **Publish failed when the stream was missing** — NATS recreated with ephemeral storage, or the reconnect handler not yet done re-ensuring it. Now recreates the stream via `ensureStream` and retries once.
- **NATS unreachable at boot disabled routing for the life of the process.** `NewNATSEventBus` failed after ~20s, the controller logged "channel routing disabled" and carried on without a bus; and a router whose initial `Subscribe` failed returned that error from `Start`, taking the whole manager down. This is the node-drain / rolling-upgrade case where the NATS pod and the controller are recreated together. The controller's bus now provisions lazily: startup returns a usable bus, and `Subscribe` returns a live channel that creates its consumer once NATS answers.

**Behavior change:** `NewNATSEventBus` no longer returns an error for an unreachable NATS (configuration errors are still returned). `NewNATSEventBusWithContext` — used by the API server, whose webproxy subscribes per request — keeps its fail-fast contract unchanged.

**Testing:** `nats_resilience_test.go` in the existing no-server style: `TestIsStreamGoneErr`, `TestConsumerConfigSurvivesFetchGaps`, `TestLazyStartupReturnsUsableBusWhenNATSUnavailable` (bus + live channel from a refused loopback port), `TestBoundedStartupStillFailsWhenNATSUnavailable`. `go test -race ./internal/eventbus/` passes.

### DIFF
--- a/internal/eventbus/nats.go
+++ b/internal/eventbus/nats.go
@@ -3,6 +3,7 @@ package eventbus
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log"
 	"time"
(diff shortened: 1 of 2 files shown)


AlexsJones (approved):
Thanks for chasing the remaining gaps from #254, the writeup makes the failure modes concrete.

Read the full diff plus the new tests. The classification in `isStreamGoneErr` is careful (consumer-not-found correctly stays a consumer problem for the existing recreate path), and keeping fail-fast semantics on the `WithContext` constructor while only the long-lived one goes lazy is the right split. The real-server regression covering both lazy startup and stream deletion is honest end-to-end proof, and the no-server tests pin both behaviors without needing infrastructure. One tradeoff is already documented in the code (orphaned consumers linger up to the hour threshold), which is the right call to surface. Ready for the maintainer's final look.

PR #575, opened 2026-09-15
Centralise isTerminal Phase determination

kristof-ringleff:
Adds `AgentRunPhase.IsTerminal()` covering Succeeded, Failed and Skipped, and routes every
terminal-phase check through it instead of comparing phases inline.

Three of the old inline checks were wrong. The WorkspaceSession controller's `isTerminalPhase`
omitted Skipped, so a skipped run kept holding the session lock and blocked peer admission.
The `workspace` CLI's `isTerminalRunPhase` omitted it too. The TUI's `/abort` compared against `"Completed"`, which
is a Sandbox CR phase and never an AgentRun phase, so it deleted succeeded runs instead of
reporting them already finished.


AlexsJones (approved):
Nice consolidation, and good catch on the three inline checks that had drifted. `AgentRunPhase.IsTerminal()` covers exactly Succeeded, Failed and Skipped, every caller now goes through it, and the fixes fall out of that: a Skipped run releases the workspace session lock, the CLI stops listing skipped runs as live, and `/abort` no longer deletes a Succeeded run because it was comparing against `"Completed"`. `isAgentRunActive` becoming `!IsTerminal()` keeps `""` active and treats any future phase as active, which is the safe default for Forbid schedules. Verify is green. Approving.

PR #581, opened 2026-09-16
fix(channel): reply on agent failure and stop cross-instance outbound…

SamuZad:
… delivery

Two channel-routing gaps:

- Failed runs produced total silence. The router subscribed to agent.run.completed but not agent.run.failed, so a timed-out or crashed run that started from Slack/Telegram never replied. Subscribe to failures and reply in the originating thread, choosing the advice from the classified reason bucket the controller already puts in the event (timeout -> point at spec.agents.default.runTimeout, token budget, policy, OOM, model unavailable) and quoting the underlying error. Attributed via the agent display name like a normal reply.

- Every channel pod received every agent's outbound messages. All pods subscribe to the single channel.message.send topic and nothing filtered by instance, so with several channel-enabled agents each pod acted on replies addressed to the others. BaseChannel.SubscribeOutbound now drops events whose instanceName is not its own.


SamuZad:
@AlexsJones yes another one for your consideration

AlexsJones (changes requested):
Thanks for both fixes, the outbound instance filter in particular closes a real duplicate-delivery hole. I checked it against every publisher: channel runs set `AgentRef` to the channel pod's own agent, and the IPC bridge and the router both stamp that name, so replies still reach the right pod.

One thing to fix before merge: failures from inside the agent would now get two replies. When the runner hits an error it writes `result.json` with `status: "error"` and exits 1 (`fatal()` in `cmd/agent-runner/main.go`). The bridge publishes that file as `agent.run.completed`, so `handleCompleted` already posts `Error: ...` to the thread. The Job then fails, `failRun` publishes `agent.run.failed`, and the new `handleFailed` posts a second message. Your `llm_error` test case is exactly that path. Simplest fix: have `handleCompleted` stay silent when `result.Status` is `error` and let `handleFailed` own every failure reply, since it has the better advice. A test that feeds both events for one run and expects a single outbound message would pin it.

SamuZad:
> Thanks for both fixes, the outbound instance filter in particular closes a real duplicate-delivery hole. I checked it against every publisher: channel runs set `AgentRef` to the channel pod's own agent, and the IPC bridge and the router both stamp that name, so replies still reach the right pod.
> 
> One thing to fix before merge: failures from inside the agent would now get two replies. When the runner hits an error it writes `result.json` with `status: "error"` and exits 1 (`fatal()` in `cmd/agent-runner/main.go`). The bridge publishes that file as `agent.run.completed`, so `handleCompleted` already posts `Error: ...` to the thread. The Job then fails, `failRun` publishes `agent.run.failed`, and the new `handleFailed` posts a second message. Your `llm_error` test case is exactly that path. Simplest fix: have `handleCompleted` stay silent when `result.Status` is `error` and let `handleFailed` own every failure reply, since it has the better advice. A test that feeds both events for one run and expects a single outbound message would pin it.

@AlexsJones thanks for the review and the suggestion. Made the change

AlexsJones (approved):
Thanks for turning that around so quickly, and `TestFailedRun_ExactlyOneReply` is exactly the test this needed. I also checked that silencing `handleCompleted` on `status: error` can't drop a failure reply. The only place that writes an error run result is the runner's error exit (`fatal()` and the `res.Status == "error"` branch in `cmd/agent-runner/main.go`), and both exit 1, so `failRun` always follows. The prompt service's error results go to `result-<id>.json`, not the run result, and the controller's gated completion always publishes `success`. Verify is green on 8e8c567. Good to merge.

SamuZad:
@AlexsJones (apologies for the repeated tags) could I get a merge on this?

PR #589, opened 2026-09-17
fix(controller): measure the postRun timeout from the postRun Job, not the agent run

kristof-ringleff:
Split out of #338 at review request — this fix stands on its own and doesn't
depend on the retry design.

## The bug

`reconcilePostRunning` measured the postRun budget from `Status.StartedAt`, the
*agent run's* start:

```go
if agentRun.Status.StartedAt != nil {
    elapsed := time.Since(agentRun.Status.StartedAt.Time)
    postRunTimeout := 10 * time.Minute
    if elapsed > postRunTimeout { /* kill the Job */ }
}
```

Any run longer than ten minutes therefore arrives in `PostRunning` with its
budget already spent, so the very first PostRunning reconcile deletes the
postRun Job. For a gated run that means the gate can never write a verdict: the
run always resolves as `timeout` and falls through to `gateDefault`. Short runs
were unaffected, which is why it went unnoticed — the two anchors agree there.

## The fix

Both bounds now measure from the postRun Job's own clock.

- **`postRunJobStart`** prefers `Job.Status.StartTime`, which is exactly what
  the Job controller uses for `ActiveDeadlineSeconds`, so the two stay
  consistent. It falls back to `CreationTimestamp` for a Job that never starts
  (stuck on quota, or no Job controller under envtest) so such a Job still times
  out rather than holding the run forever.

- **`postRunBudget`** honours `LifecycleHookContainer.Timeout` — a field that
  has been documented and served in the CRD but never read. PostRun hooks run
  sequentially as init containers, so the budget is the **sum** of the declared
  timeouts, 5m each when unset, floored at 10m. The floor means a spec that
  declares no timeout keeps exactly the budget it has today; nothing gets
  tighter. The budget now drives the Job's `ActiveDeadlineSeconds` too, and the
  controller-side backstop sits 30s behind it so the Job's own deadline fires
  first and the failure arrives through the ordinary `Failed` path instead of
  the controller racing it to a delete.

Kubernetes has no per-init-container timeout, so `timeout` bounds the Job as a
whole rather than each container — a hook that overruns eats into what is left
for the hooks after it. The docs say so explicitly.

A human-in-the-loop gate can now declare `timeout: 24h` and actually get it,
which is the case the `requireApproval` toggle in the API/UI depends on.

## No CRD change

`LifecycleHookContainer.Timeout` already exists in `api/v1alpha1` and is already
in the generated CRDs. This PR only starts reading it, so `make manifests` is a
no-op.

## Tests

`internal/controller/agentrun_postrun_timeout_test.go` covers the anchor, the
budget table, and four `reconcilePostRunning` cases: a 2h run whose gate has
barely started (the regression), a genuinely stale gate Job that must still be
killed, a 24h declared timeout surviving past the default, and a short run left
unchanged. Reverting the `reconcilePostRunning` hunk fails the first and third;
both pass with the fix.

`go build ./...`, `go vet ./...`, `gofmt -l` and `go test -race
./internal/controller/` are clean. The only failure in a full `go test -race
./...` is `internal/cellnparent`'s
`TestRegisteredParentBindsLiveSelectionWithoutWidening`, which fails the same
way on an unmodified `main` here.

https://claude.ai/code/session_01LUgAdk6mwW2TecDbMtchF8


AlexsJones (approved):
Thanks for splitting this out, it reads much better on its own. Measuring from `postRunJobStart` and driving `ActiveDeadlineSeconds` from the same `postRunBudget` fixes the long-run gate kill. The 10m floor keeps existing specs unchanged, and the 30s grace lets the Job's own deadline fail first. The 2h-run-with-a-fresh-gate test pins the regression.

One non-blocking follow-up: `applyRequireApproval` returns early when `manual-approval-gate` is already present. So agents that enabled approval before this change keep a hook with no `timeout`, and still get 10 minutes until the toggle is turned off and on again. Worth a release note or a small migration. Approving, CI is green.

AlexsJones (approved):
Nice, the new `Timeout` description is exactly what a user needs: hook timeouts are summed into one Job deadline, floored at 10 minutes, and preRun hooks don't honour it yet. It matches `postRunBudget` and `postRunMinTimeout` in the controller, and the regenerated CRDs agree with the Go comment. Verify and the envtest suite are green at `e76c2a58`. Still approving.

PR #590, opened 2026-09-17
fix(ci): run the system tests in verify

kristof-ringleff:
Nothing in CI runs `make test-system`, so test/system has been broken without anyone noticing. Add it to the verify job, and pin setup-envtest alongside controller-gen so a new release can't break CI without a change here.

The first run is expected to fail: the suite does not currently compile.

The second commit fixes the test for the updated Handler to take `*tokenReader` into account. Additionally removed the cached read on the api server to remove test flakiness.


AlexsJones (approved):
Good catch that nothing ran `make test-system`. Adding it to verify and pinning setup-envtest to v0.25.1 both make sense. So do the suite fixes: an uncached client to stop the create-then-read races, and `Handler(nil)` for the no-auth path. Verify is green at `c4a4a1b` with the system tests running. Approving.

kristof-ringleff:
I have removed the extra github workflow step since recent changes already included these - no need to run them twice. This is now reduced to the cached read changes in the test to avoid cache races during tests.

AlexsJones (approved):
Thanks for keeping this current after the CI half landed on main separately. What's left still stands on its own: pinning `setup-envtest` to v0.25.1 instead of `@latest`, and the uncached client in `test/system/suite_test.go` so create-then-read assertions stop racing the informer. Verify and System tests (envtest) are both green at `0e127d8b`. Approving.

